PR #47, opened 2026-04-24
test: strengthen GDN coherence check + document recurrent-state collapse regression

kekzl:
## Summary
- **Test**: strengthen \`GDNModelTest.GenerateCoherentOutput\` from a size-only check to a token-repetition-ratio check (unique / total ≥ 30 %, fires only when output has ≥10 words).
- **TODO.md**: document the Qwen3.5-4B/9B Q8_0 recurrent-state-collapse regression that reproduces on clean \`imp:main\` today.

## Background
The existing assertion was \`text.size() > 5\`, which trivially passes on degenerate repetition like \`"my my my my my my my my my my"\`. During a full audit today I reproduced this exact failure on clean main (no Phase-4 changes), on both Qwen3.5-4B and Qwen3.5-9B Q8_0, and across \`IMP_GDN_REF=1\` / \`IMP_DEBUG_RAW=1\` / various sampling configs. Qwen3-4B (dense) does not reproduce, isolating the bug to the GDN scan path.

The test now catches this bug class. The bug itself needs layer-diff-vs-llama.cpp to find the root cause (separate work, tracked in TODO.md).

## Test plan
- [x] Rebuild with \`IMP_BUILD_TESTS=ON\`; full unit suite still passes (test is gated on \`IMP_TEST_MODEL_GDN\`)
- [x] With \`IMP_TEST_MODEL_GDN=/models/Qwen3.5-4B-Q8_0.gguf\`: the "Write a paragraph about Jupiter" prompt happens to echo the input deterministically (100 % unique), so the test passes — but severe \`my my my...\` cases would now fail

## Does NOT fix the underlying bug
The GDN recurrent-state collapse is pre-existing on main and not related to my in-flight Phase-4 refactor PRs (#40-#46). Fixing it is tracked separately in TODO.md as the top open issue.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -2,6 +2,36 @@
 
 ## Open Work
 
+### Qwen 3.5 GDN Recurrent-State Collapse (REGRESSED from v0.6)
+**Severity: high — blocks production on all Qwen3.5 Q8_0 models.**
+
+Raw-completion and chatml generation both collapse into token repetition after 3–8 tokens on Qwen3.5-4B and Qwen3.5-9B Q8_0:
+
+```
+prompt: "Hello world, my name is"
+output: " my my my my my my my..."
+
+prompt: "Once upon a time in a faraway kingdom..."
+output: " journeyed a time who whoed a time."
+```
+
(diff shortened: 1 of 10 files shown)


kekzl:
Replacing with clean branch test/gdn-coherence-fresh off main (original branched off phase-4 stack by mistake)

PR #62, opened 2026-04-25
runtime: drop FP16 fallback for GDN MXFP4 + Qwen3.5-27B QK-norm split

kekzl:
## Summary

Three coupled changes that together let Qwen3.5-27B-mxfp4 load on a 32 GiB GPU without VRAM oversubscription. The model still doesn't produce fully coherent output (separate architecture-conversion mismatch), but it loads, runs, and stops cascading IMAs.

User observation that motivated the work: **\"79 GB VRAM\"** for what should be a 12 GiB MXFP4 model. The old code path forced GDN MXFP4 models through a 48 GiB FP16 fallback on top of the 12 GiB MXFP4 raw, total 60+ GiB on a 32 GiB GPU.

## Changes

1. **executor_pre_dequant.cu** — Drop the historical \"GDN models need FP16 fallback\" exclusion. Every linear projection in `executor_ssm_gdn.cu` now goes through `gemm_dispatch` which routes to the MXFP4 path correctly; the scan kernel only touches FP32 tensors (A_log, dt_bias, conv1d, ssm_norm) which are never MXFP4.

2. **attention_cublas.cu** — Use the FP32 S-matrix whenever the scratch buffer fits, not only when `scale==1.0` (Gemma-4). The FP16 S path can accumulate enough round-off across deep layers to NaN at attention layers in the second half of the model.

3. **executor_attention.cu** — Two parts:
   - Pass the full `attn_scores_` tensor (capacity = max seq_len²) to `attention_cublas_prefill` instead of a sub-view sized for the current n. The FP32-fits check was reading the view's shape, not the buffer's actual capacity.
   - Detect `attn_q_norm`/`attn_k_norm` with norm_dim < head_dim and reshape Q/K accordingly. Qwen3.5-27B-mxfp4 ships with a single (256,) norm weight intended to be applied per 256-dim chunk along the head_dim=512 Q. The previous code read 256 elements past the end of the weight buffer and produced the L59 attention NaN that poisoned the rest of the forward pass.

## Test plan

- [x] Qwen3.5-4B Q8_0: coherent output (tg=161 tok/s)
- [x] Qwen3.5-9B Q8_0: coherent output (\"in a magical land called the Math Kingdom, there lived a group of\")
- [x] Qwen3.6-35B Q4_K_M: still works
- [x] Qwen3.5-27B-mxfp4: VRAM 23 GiB (was 60+, oversubscribed). No more L59 NaN cascade.
- [x] `imp-tests`: 81/94 pass, 13 skipped, 0 fail.

## Remaining work

Qwen3.5-27B-mxfp4 still produces incoherent output. The L59 IMA is fixed and the residual stream stays finite end-to-end, but the actual logits don't match what the model is trained to emit. Suspected cause: another aspect of the GGUF format that this converter uses differently from the Q8_0 variants (`attn_q_norm` having mixed-sign values and shape (256,) is unusual; the `ssm_dt.weight` → A_log mapping in PR #61 is one example we *did* find).


kekzl:
Superseded by #66 — only the two cleanly-applicable changes from `569af55` (FP32 attention S-matrix + Qwen3.5 QK-norm split) plus `b56af82` (tools cleanup) made it across; the `executor_pre_dequant.cu` GDN→FP16 fallback drop is deliberately skipped because main has explicit diagnostic comment from #60 noting the cuBLAS-INTERNAL_ERROR cascade for GDN-shape MXFP4 weights is still unrooted, and per the original commit message the model was incoherent anyway.

PR #154, opened 2026-05-09
perf(gdn): 4-way input fusion (ssm_in + gdn_gate + alpha + beta) for decode

kekzl:
## Summary

Step 2 of the GDN GEMV-fusion series (Step 1 = #153, alpha+beta only). The four input projections that share `no` as input — `ssm_in`, `gdn_gate`, `gdn_alpha`, `gdn_beta` — are concatenated along N (output dim) into one `[conv_channels + inner + 2*n_heads, d_model]` packed weight built at load time. For n=1 decode the engine runs a single `gemm_dispatch` and takes offset views for proj / gate_out / alpha / beta — no copies, no deinterleave. Prefill (n>1) keeps the original four-call dispatcher unchanged; the originals stay on-device, so this is purely additive.

## Bench

RTX 5090, identical seed/prompt, 256-token decode, graphs ON:

| Model                    | clean main | Step 1 | Step 2 | Δ vs main |
|--------------------------|-----------:|-------:|-------:|----------:|
| Qwen3.6-35B-A3B-NVFP4    | 171.20     | 173.28 | **178.0** | **+4.0 %** |
| Qwen3.5-4B-GDN Q8_0      | 193.45     | 193.83 | 193.7  | unchanged |
| Qwen3-4B Q8_0 (dense)    | (`make verify-fast` +/- noise; fusion inactive on dense models) |

`make verify-fast` PASS on Qwen3-8B Q8 (canonical model — fusion inactive there) once the GPU is warm.

## Buffer routing

The fused output lives in a new `gdn_fused_proj_buf_` workspace tensor sized only on `has_gdn_` models. `ssm_proj_buf_` stays sized for the SSM/Mamba2 `ssm_in_dim` and is unaffected — conv1d's FP32 output still overwrites it as before, while gate / alpha / beta slices in `gdn_fused_proj_buf_` remain valid through the rest of the layer.

## Fallback chain in `run_gdn()`

1. `n==1 && ly.gdn_input_packed.data`   → 4-way fused path (this PR)
2. `n==1 && ly.gdn_alpha_beta_packed.data` → Step 1 alpha+beta fusion (PR #153)
3. otherwise → four separate `gemm_dispatch` calls (original)

## Filter

4-way pack only fires when all four weights are FP16/BF16 on-device with matching `d_model`. NVFP4-prequant Qwen3.5/3.6 keeps SSM internals at FP16 (excluded from the NVFP4 cache for accuracy — see `weight_map.cpp:842`), so the pack lights up there. Q*_K / MXFP4 paths still get Step 1's alpha+beta-only fallback.

## VRAM cost

Per GDN layer ~32 MiB extra (the packed buffer is essentially a duplicate of the four input-projection weights). Qwen3.6-35B with 60 GDN layers → ~1.9 GiB extra. On 32 GiB cards this leaves ~4-5 GiB for KV cache (≈8K context). For tighter budgets we keep the originals so the executor already has a fallback — disabling the pack via env var is a trivial follow-up if it becomes annoying.

## Out of scope

- **Prefill (`n>1`) fusion**: needs deinterleave of the `[n, total_out]` output into the existing per-slice buffers, plus conv1d in/out buffer swap because in-place compaction has token-overlap races. Maintainable but bigger; separate PR.
- **Free originals**: paired with prefill-fused work above to make the pack VRAM-neutral instead of additive.

## Test plan

- [x] `make verify-fast` (build + tests + perf + smoke + graphs gate)
- [x] Qwen3.6-NVFP4 4-run bench: 178.18 / 178.27 / 176.93 / 177.92 → median 178.0 vs 171.2 baseline = +4.0 %
- [x] Qwen3.5-GDN Q8_0 3-run bench: 193.19 / 193.96 / 194.07 → no change (4-call path used; Q8_0 weights skip the FP16/BF16 filter)
- [x] Qwen3.6-NVFP4 coherent decode (\"capital of France\" smoke)
- [x] Qwen3.5-GDN Q8_0 coherent decode (Babbage essay)

### DIFF
--- a/src/graph/executor.h
+++ b/src/graph/executor.h
@@ -567,6 +567,9 @@ private:
     Tensor ssm_z_buf_;     // [max_tokens, inner_size] for gate
     Tensor ssm_out_buf_;   // [max_tokens, d_model] for ssm_out projection
     Tensor ssm_dt_buf_;    // [max_tokens, n_heads] for dt after split
+    Tensor gdn_fused_proj_buf_;  // [max_tokens, conv_channels+inner+2*n_heads] FP16 — output
+                                 // of the fused GDN input GEMV when ly.gdn_input_packed is
+                                 // built; sized only for has_gdn_ models.
 
     // --- Separately allocated buffers (not part of unified workspace) ---
 
--- a/src/graph/executor_ssm_gdn.cu
+++ b/src/graph/executor_ssm_gdn.cu
@@ -279,12 +279,37 @@ void GraphExecutor::run_gdn(int layer, const InferenceState& state, cudaStream_t
     IMP_CUDA_CHECK_LOG(cudaMemcpyAsync(r.data, h.data, h.nbytes(), cudaMemcpyDeviceToDevice, stream));
     rmsnorm(h, ly.attn_norm, no, eps, stream, norm_w_off_);
 
+    // M=1 decode 4-way input fusion: when the load-time pack succeeded, run a
+    // single GEMV against [conv_channels+inner+2*n_heads, d_model] and slice
+    // the output for proj / gate_out / alpha / beta. Saves 3 gemv launches per
+    // layer per decode step. Prefill (n>1) keeps the 4-call path unchanged so
+    // we don't have to deinterleave the GEMM output.
+    const bool fused_input = (n == 1) && (ly.gdn_input_packed.data != nullptr) &&
+                             (gdn_fused_proj_buf_.data != nullptr);
+    int packed_conv_channels = fused_input ? ly.gdn_packed_conv_channels : 0;
+    int packed_inner = fused_input ? ly.gdn_packed_inner : 0;
+    int packed_n_heads = fused_input ? ly.gdn_packed_n_heads : 0;
+
     // 2. ssm_in (attn_qkv) projection → [n, conv_channels]
     //    GDN: no z-split, no dt — the full projection goes to conv1d.
-    //    ssm_proj_buf_ is [max_tokens, ssm_in_dim] but we only need [n, conv_channels].
     int64_t proj_shape[2] = {static_cast<int64_t>(n), static_cast<int64_t>(conv_channels)};
-    Tensor proj(ssm_proj_buf_.data, compute_dtype_, 2, proj_shape, true);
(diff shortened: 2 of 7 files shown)


kekzl:
## Critical bugfix — root cause found

The original Step 2 PR had a workspace-sizing bug that caused the IMA crash @kekzl reproduced via Open WebUI on Qwen3.6-NVFP4. Just shipped \`26b879a\` on this branch.

**Root cause:** `executor_workspace.cu` initialised `has_gdn_ = false` and ran `compute_shared_sizes()` *before* the GDN detection scan that flips it to true. So `compute_shared_sizes()` undersized **three** SSM buffers, while `configure_ssm_workspace()` (running with `has_gdn_ = true`) laid them out at the correct full size:

| Buffer | undersized as | should be | delta |
|---|---|---|---|
| `ssm_proj_buf_` | FP16 (es=2) | FP32 (sizeof(float)) | −48 MiB |
| `ssm_dt_buf_` | 1× n_heads | 2× (alpha+beta) | −0.13 MiB |
| `gdn_fused_proj_buf_` | skipped | full layout | −48 MiB |

The shared_workspace_ malloc honoured the (smaller) computed size, while configure advanced the pointer for the (correct) larger layout. Net: the last ~96 MiB of SSM writes landed *past* the end of `shared_workspace_`, scribbling on whatever cudaMalloc placed adjacent.

**Symptom progression:** Qwen3.6-NVFP4 ships GDN weights as BF16 (verified in safetensors header), so the 4-way pack lit up for 30 layers. Multi-turn ran fine for ~5 turns, then a hybrid-arch cancel left the engine in a state where the next decode step's `cudaMemcpyAsync(&h_result, d_result, …)` (sampling.cu:880) hit IMA. The error cascaded through every later cudaMemcpy until the request returned 600 tokens of garbage.

**Validation:**

| | Shared workspace | 10-turn stress (long generations + hybrid-arch cancels) |
|---|---:|---|
| pre-fix | 120.2 MiB ssm | IMA after 4 turns, cascade |
| post-fix | 216.75 MiB ssm | zero IMA, 4 successful generations + 6 clean cancels |

Same code change makes the original measured perf win (+4.0% Qwen3.6-NVFP4 decode) safe to ship. PR ready for re-review.

PR #259, opened 2026-05-18
perf(q4k-imma): Phase 2B.3 — BLOCK_M=64 N=32, WRM·WRN=2·2 per warp

kekzl:
## What

Expands the Phase 2B.2 multi-warp INT8 IMMA kernel to a larger output tile per CTA. Each warp now owns a 32×16 sub-tile (4 m16n8k32 fragments), iterated as a 2×2 grid of (wrm, wrn) MMAs per K-block. **16 MMAs per CTA per K-block, 4× Phase 2B.2's 4.**

## Why

Phase 2B.2 (PR #258) plateaued at ~19 TOPS due to limited MMAs-per-CTA. The CTA undersaturation analysis showed multi-warp alone insufficient; expanding output-per-CTA was the design memo's next step.

## How

| | Phase 2B.2 | Phase 2B.3 |
|---|---|---|
| BLOCK_M × BLOCK_N | 32 × 16 | **64 × 32** |
| Warps / CTA | 4 (2×2 spatial) | 4 (2×2 spatial) |
| MMAs per warp per K-block | 1 | **4** (WRM·WRN=2·2) |
| MMAs per CTA per K-block | 4 | **16** |
| SMEM per CTA | 3072 B | 6144 B |
| Load distribution | 64 A + 64 B (split) | All 128 do both (1 ca_16 + 1 ca_8 each) |

Same 2-stage cp.async pipeline carries over. Tests updated to BLOCK_M=64 N=32 multiples (the 32×16 sizes fall below the new divisibility guard).

## Tests

All 4 correctness tests PASS at FP16 ulp-floor accuracy:

| Test          | Shape           | max_abs |
|---            | ---             | ---:    |
| Tiny          | 64 × 32 × 32    | 0.37 |
| MultiSub      | 64 × 32 × 256   | 1.00 |
| MultiTile     | 128 × 64 × 128  | 0.96 |
| FFNLikeShape  | 128 × 64 × 512  | 1.70 |

Max rel_err ~0.05 % across all shapes — math identity holds.

## Bench

Production FFN-scale shapes (Qwen3-32B / Gemma-3-12B style), mean of 20 reps, K=2048:

| M    | N    | ms/rep | TOPS |
|---:  |---:  |   ---: | ---:|
| 512  | 512  | 0.058  | 18.7 |
| 1024 | 512  | 0.071  | 30.4 |
| 2048 | 512  | 0.139  | 31.0 |
| 4096 | 1024 | 0.438  | 39.3 |
[...]

PR #260, opened 2026-05-18
perf(q4k-imma): Phase 2B.4 — 3-stage cp.async pipeline (mixed result)

kekzl:
## What

Deepens the Phase 2B.3 cp.async pipeline from 2 stages to 3 on top of the same BLOCK_M=64 N=32, WRM·WRN=2·2 tile. Prologue now commits kb=0 and kb=1; loop prefetches kb+2 while waiting on kb. SMEM grows to \`sA[3][64][32]\` + \`sB[3][32][32]\` = 9216 bytes per CTA.

Tail-handling branches between \`wait_group<2>\` (steady state), \`wait<1>\` (no new prefetch but kb+1 in flight), and \`wait<0>\` (last iter).

## Why

Phase 2B.3 plateaued at ~40 TOPS. The 2-stage pipeline may be too shallow to hide all HBM latency at long K. This PR tests whether deeper pipelining moves the plateau.

## Tests

All 4 correctness tests still PASS at FP16 ulp-floor accuracy.

## Bench

Production FFN shapes, mean of 20 reps, K=2048:

| M    | N    | 2B.3 (2-stage) | 2B.4 (3-stage) | Δ      |
|---:  |---:  |            ---:|            ---:| ---:   |
| 512  | 512  | 18.7 TOPS      | 25.2 TOPS      | +35 %  |
| 1024 | 512  | 30.4           | 37.7           | +24 %  |
| 2048 | 512  | 31.0           | 33.9           | +10 %  |
| 4096 | 1024 | 39.3           | 40.6           | +3 %   |
| 2048 | 2048 | 40.8           | 38.0           | -7 %   |
| 4096 | 4096 | 40.1           | 38.5           | -4 %   |

**Mixed result.** The third stage helps small/medium shapes where per-CTA latency was the bottleneck but slightly regresses at full-saturation shapes where the extra SMEM + longer prologue overhead doesn't amortise. The **~40 TOPS plateau is firmly the architecture ceiling** for this kernel design; 2-stage vs 3-stage is a per-shape trade-off a Phase 2C dispatcher should resolve.

## Risk

None. Standalone bench kernel.

## What's left for the next phase

The real next perf lever is **`ldmatrix.x4`** (Phase 2B.5) — closes the SMEM→reg bandwidth gap that's now the binding constraint at large M/N. Past the ldmatrix step, fundamentally different tile shapes (warp-group MMA, persistent CTAs) would be needed to approach the 931 TOPS raw ceiling — CUTLASS-territory restructures, not incremental tweaks.

## Stacking note

This PR is stacked on \`topic/q4k-imma-phase2b3-bigtile\` (#259). Both target \`main\` directly; merge order doesn't matter.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -200,6 +200,7 @@ if(IMP_USE_CUTLASS)
         list(APPEND IMP_COMPUTE_SOURCES src/compute/mxf4nvf4_qkt_validate.cu)
         list(APPEND IMP_COMPUTE_SOURCES tests/bench/tma_block_scale_bench.cu)
         list(APPEND IMP_COMPUTE_SOURCES tests/bench/fmha_v_load_bench.cu)
+        list(APPEND IMP_COMPUTE_SOURCES tests/bench/mmq_q4k_imma_tile_bench.cu)
     endif()
     if(IMP_BUILD_TESTS)
         set_source_files_properties(src/compute/gemm_grouped_nvfp4_smallM.cu
@@ -477,6 +478,7 @@ if(IMP_BUILD_TESTS)
         tests/test_mxf4nvf4_qkt_validate.cu
         tests/test_tma_block_scale_bench.cu
         tests/test_fmha_v_load_bench.cu
+        tests/test_mmq_q4k_imma_tile_bench.cu
         tests/test_weight_dispatch.cu
     )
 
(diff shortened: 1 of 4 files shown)


kekzl:
Refuted at saturation (mixed result: +35% small M, −7% saturation). Not landing — findings preserved in #267's commit + Phase 2B ceiling memo (#262).

PR #261, opened 2026-05-18
perf(q4k-imma): Phase 2B.5 — ldmatrix.x4/x2 SMEM→reg (neutral, architecture ceiling)

kekzl:
## What

Replaces the manual 4×uint32 / 2×uint32 SMEM loads for the A and B fragments with PTX \`ldmatrix.sync.aligned.m8n8.x4.shared.b16\` and \`.x2\` — one PTX instruction loads the full m16n8k32 fragment in the canonical register layout per lane, with the warp's 32 lanes providing row addresses for the 4 (\`.x4\`) or 2 (\`.x2\`) 8×8-b16 sub-matrices.

## Why

Phase 2B.4 plateaued at ~38–40 TOPS. The hypothesis was that SMEM→register bandwidth was the remaining bottleneck; \`ldmatrix\` is the canonical fix.

## How

Address mapping (PTX ISA 8.5 §9.7.13.6, \`.s8.s8.s32\` m16n8k32 case):

| Operand | Matrix | Lane group | Rows / cols |
|---      | ---:   | ---        | ---         |
| A.x4 | 0 | 0..7   | rows 0..7,  K 0..15 |
|      | 1 | 8..15  | rows 8..15, K 0..15 |
|      | 2 | 16..23 | rows 0..7,  K 16..31 |
|      | 3 | 24..31 | rows 8..15, K 16..31 |
| B.x2 | 0 | 0..7   | cols 0..7,  K 0..15 |
|      | 1 | 8..15  | cols 0..7,  K 16..31 |

Lanes 16..31 for \`.x2\` are inactive (PTX silently ignores their addresses).

## Tests

All 4 correctness tests PASS at **bit-identical** max_abs values to Phase 2B.4 (0.37 / 1.00 / 0.96 / 1.70). Implementation correctness confirmed.

## Bench

Production FFN shapes, mean of 20 reps, K=2048:

| M    | N    | 2B.4 (no ldmat) | 2B.5 (ldmat) | Δ      |
|---:  |---:  |             ---:|          ---:| ---:   |
| 512  | 512  | 25.2 TOPS       | 11.7 TOPS    | -54 %  |
| 1024 | 512  | 37.7            | 35.8         | -5 %   |
| 2048 | 512  | 33.9            | 31.6         | -7 %   |
| 4096 | 1024 | 40.6            | 39.9         | -2 %   |
| 2048 | 2048 | 38.0            | 38.4         | +1 %   |
| 4096 | 4096 | 38.5            | 37.5         | -3 %   |

**Neutral-to-negative result.** \`ldmatrix\` is correct but does not move throughput at saturation. The −54 % at the smallest shape is a launch-overhead artifact; the small-M tail is dispatcher territory.

## What this proves

SMEM→reg bandwidth was **not** the binding constraint at the 38–40 TOPS plateau. Remaining bottlenecks are elsewhere — most likely scale-apply FMAs (16 per warp per K-block) or per-K \`__syncthreads\` overhead at the SM scheduler.

**This is the architecture ceiling** for the hand-rolled tile approach. The kernel as it stands (Phase 2B.3 with 2-stage cp.async, ~40 TOPS) is the recommended Phase 2C production target — already competitive with the dequant→cuBLAS path the FFN cache-miss workload would hit, even if it doesn't beat direct cuBLAS FP16-TC.

Closing the gap to cuBLAS FP16-TC (~244 TFLOPS) requires fundamentally different tile shape — warp-group MMA (tcgen05.\* on SM100), persistent-CTA with stream-K scheduling, or CUTLASS template instantiation. **Multi-week separate kernel projects**, not incremental tweaks.

## Risk

None. Standalone bench kernel.

## Stacking note

This PR is stacked on \`topic/q4k-imma-phase2b4-3stage\` (PR #260). Both target \`main\` directly; merge order doesn't matter.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -200,6 +200,7 @@ if(IMP_USE_CUTLASS)
         list(APPEND IMP_COMPUTE_SOURCES src/compute/mxf4nvf4_qkt_validate.cu)
         list(APPEND IMP_COMPUTE_SOURCES tests/bench/tma_block_scale_bench.cu)
         list(APPEND IMP_COMPUTE_SOURCES tests/bench/fmha_v_load_bench.cu)
+        list(APPEND IMP_COMPUTE_SOURCES tests/bench/mmq_q4k_imma_tile_bench.cu)
     endif()
     if(IMP_BUILD_TESTS)
         set_source_files_properties(src/compute/gemm_grouped_nvfp4_smallM.cu
@@ -477,6 +478,7 @@ if(IMP_BUILD_TESTS)
         tests/test_mxf4nvf4_qkt_validate.cu
         tests/test_tma_block_scale_bench.cu
         tests/test_fmha_v_load_bench.cu
+        tests/test_mmq_q4k_imma_tile_bench.cu
         tests/test_weight_dispatch.cu
     )
 
--- a/tests/bench/mmq_q4k_imma_tile_bench.cu
+++ b/tests/bench/mmq_q4k_imma_tile_bench.cu
@@ -41,7 +41,7 @@ namespace {
 constexpr int kBlockM = 64;
 constexpr int kBlockN = 32;
 constexpr int kBlockK = 32;
-constexpr int kNumStages = 2;
+constexpr int kNumStages = 3;  // 3-stage cp.async pipeline (Phase 2B.4)
 constexpr int kNumWarps = 4;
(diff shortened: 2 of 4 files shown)


kekzl:
Refuted: bit-identical output, neutral-to-negative perf — proves SMEM bandwidth was not the binding constraint. Not landing — findings preserved in #267 + Phase 2B ceiling memo (#262).

PR #422, opened 2026-05-25
perf: full FP8 prefill budget — pp512 50.8k tok/s, beats vLLM by 53%

kekzl:
Use all available VRAM for FP8 cache instead of conservative cap. 252/252 tensors cached. pp512: 24k→50.8k (+112%). pp2048: 26k→48k (+85%). Now faster than vLLM on all metrics.


kekzl:
Superseded by clean rebased PR

PR #468, opened 2026-05-29
feat(nvfp4-moe): fast per-expert NVFP4 MoE decode (+54-84% decode, matches/beats llama.cpp)

kekzl:
## What

Native-NVFP4 (Modelopt/llm-compressor) MoE experts ran **M=1 decode through the CUTLASS grouped-GEMM** (the prefill path), which under-utilizes the GPU at batch=1. The fast per-expert `gemv_nvfp4_moe_*` kernels already existed but were unused because building their packed cache was believed to need a ~15 GiB contiguous duplicate.

**Key realization:** `convert_nvfp4_to_cutlass` *borrows* the packed data pointer — so the per-expert NVFP4 weight data is already resident **and contiguous** in VRAM (verified by diagnostic). So `cache_moe_native_nvfp4` now **borrows** that contiguous data zero-copy, copies only the small per-row micro-scales (~1/16 of weights) into a contiguous buffer when not already contiguous, plus a tiny per-expert `tensor_scales` array, and points an `NvFP4MoEQuantResult` at the existing data. This engages the existing validated `gemv_nvfp4_moe` decode kernels; prefill stays on CUTLASS.

Adds a `borrowed` flag to `NvFP4MoEQuantResult` so `free_nvfp4_moe_result` skips `cudaFree` on borrowed / VRAMAllocator-owned pointers (fixes a teardown double-free). Config `gemm.nvfp4_moe_decode` (default on); `IMP_NO_NVFP4_MOE_DECODE=1` disables.

## Measured (RTX 5090, pp512, this machine) — decode tok/s before → after

| Model | before | after | Δ | vs llama.cpp Q4_K_M |
|---|---|---|---|---|
| Qwen3-30B-A3B NVFP4 | 170.7 | **307.5** | +80% | ≈ 317 (was far behind) |
| Qwen3-Coder-30B NVFP4 | 170.9 | **307.8** | +80% | — |
| Gemma-4-26B NVFP4 | 160.2 | **258.1** | +61% | **beats 212 by +22%** |
| Qwen3.6-35B-A3B hybrid | 150.2 | **228.5** | +52% | **matches 229** (was −31%) |

Finding: the MoE expert decode (not just the FP16 SSM) was the dominant bottleneck on the GDN hybrid.

## Correctness

- Greedy output **token-for-token identical** with flag on (gemv) vs off (CUTLASS) on Qwen3-30B — numerically equivalent path.
- All models coherent incl. the GDN hybrid (state intact) and Gemma-4 ("Paris"); error scan clean.
- Full GPU test suite green; `verify-fast` green; flag A/B clean (307 vs 167 tok/s); prefill and dense/non-MoE models unaffected; no OOM.

### DIFF
--- a/docs/MISSION_JOURNAL.md
+++ b/docs/MISSION_JOURNAL.md
@@ -140,6 +140,30 @@ product gap is "make NVFP4 (recommended path) ≥ GGUF speed" (LEAD-1/2) + prefi
 
 ## Log (timestamped, append-only)
 
+### 2026-05-29 — Iteration 3: LEAD-2 LANDED (NVFP4 MoE decode, zero-copy fast path) ✅
+The "15 GiB blocker" was a misread (data is borrowed by CUTLASS = already resident). Diagnostic
+proved per-expert NVFP4 packed DATA is contiguous in VRAM; native scales resident but NOT
+contiguous. Fix: `cache_moe_native_nvfp4` now BORROWS the contiguous expert data (zero copy) +
+copies only the small scales (~1/16) into a contiguous buffer + a tiny per-expert tensor_scales
+array, building an NvFP4MoEQuantResult that points at the existing data. This engages the existing
+fast `gemv_nvfp4_moe_*` decode kernels (base+stride) instead of the CUTLASS grouped GEMM (which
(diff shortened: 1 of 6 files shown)


kekzl:
Superseded by a rebased branch (force-push unavailable in this environment) — reopening from mission/nvfp4-moe-decode-v2.

PR #489, opened 2026-05-30
feat(vision): Gemma-3 + Gemma-4 VL image understanding

kekzl:
## Problem

imp's `--mmproj`/`--image` VL path **ran end-to-end but produced garbage**: a cat photo was described as *"a complex, repeating pattern of the Hindi letter त"*. The pipeline loaded, encoded, and decoded — it just hallucinated, because the image embeddings were never projected into LLM space.

## Root cause

`src/vision/vision_loader.cpp` only recognized **LLaVA-style** projector tensor names (`mm.0.weight`, `mm.pre_norm.weight`). Gemma-3 mmproj GGUFs (`clip.projector_type = gemma3`) name them differently:

- `mm.input_projection.weight` (SigLIP 1152 → LLM d_model)
- `mm.soft_emb_norm.weight` (pre-projection RMSNorm)

Both stayed **unassigned** (`437/439` tensors), so `mm_proj_w` was **NULL** → `vision_gemm` read a null pointer → SigLIP features (1152-dim) were never projected → garbage embeddings → hallucination.

Additionally, `mm.input_projection.weight` is stored **transposed** vs the LLaVA convention: GGUF `ne=[out, in]` (`ne0=out`), whereas `vision_gemm`'s `B=[N=out, K=in]` contract (matched by `ffn_up` etc.) expects `ne0=in`.

## Fix

- Recognize both Gemma-3 projector tensor names.
- Upload `mm.input_projection.weight` via a new host-side transpose helper so its layout matches `vision_gemm`; derive `lm_d_model` from its out-dim.
- Map `mm.soft_emb_norm.weight` → the pre-projection RMSNorm (encoder already applies it as plain-mul RMSNorm, eps 1e-6 — matches llama.cpp's gemma3 clip path).

Everything else in the encoder (avg-pool 4×4 → soft_emb_norm → input_projection, 0.5/0.5 normalization, GELU) was already correct.

## Verification

`gemma-3-4b-it Q4_K_M` + `mmproj-F16`, greedy (temp 0):

| Before | After |
|---|---|
| `437/439` tensors, `mm_proj_w=NULL` | `439/439`, `LLM d_model=2560 (from mm_proj)` |
| cat → *"repeating pattern of the Hindi letter त"* | cat → *"a tabby cat … brown and gray striped coat"* |
| | pizza → *"Neapolitan pizza … tomato sauce, mozzarella, basil"* |

Different images yield distinct, accurately grounded descriptions. No degeneration, stderr clean. `make verify-fast` OK.

### DIFF
--- a/src/vision/vision_loader.cpp
+++ b/src/vision/vision_loader.cpp
@@ -248,6 +248,53 @@ bool upload_tensor_fp16(const void* src, GgufWireType type, int64_t n_elements, 
     return true;
 }
 
+// Upload a 2D weight transposed: `src` is row-major [rows, cols]; the device
+// result is row-major [cols, rows]. Gemma-3's mm.input_projection.weight is
+// stored with ne0=out (the transpose of vision_gemm's [N=out, K=in] contract,
+// where the LLaVA-style projectors store ne0=in). Transposing at load lets the
+// same vision_gemm projection path consume it unchanged.
+bool upload_tensor_fp16_transposed(const void* src, GgufWireType type, int64_t rows, int64_t cols,
+                                   void** d_out, std::vector<void*>& gpu_allocs) {
+    int64_t n = rows * cols;
+    std::vector<half> h_src(static_cast<size_t>(n));
+    if (type == GgufWireType::F16) {
+        std::memcpy(h_src.data(), src, static_cast<size_t>(n) * sizeof(half));
+    } else if (type == GgufWireType::F32) {
+        const float* f32 = static_cast<const float*>(src);
+        for (int64_t i = 0; i < n; i++)
+            h_src[i] = __float2half(f32[i]);
+    } else if (type == GgufWireType::BF16) {
+        const uint16_t* bf16 = static_cast<const uint16_t*>(src);
+        for (int64_t i = 0; i < n; i++) {
+            uint32_t bits = static_cast<uint32_t>(bf16[i]) << 16;
+            float f;
+            std::memcpy(&f, &bits, sizeof(float));
+            h_src[i] = __float2half(f);
+        }
+    } else {
+        IMP_LOG_ERROR("Vision: unsupported GGML type %u for transposed upload",
+                      static_cast<uint32_t>(type));
+        return false;
+    }
+
+    std::vector<half> h_dst(static_cast<size_t>(n));
+    for (int64_t r = 0; r < rows; r++)
+        for (int64_t c = 0; c < cols; c++)
+            h_dst[c * rows + r] = h_src[r * cols + c];
+
+    half* d_ptr = nullptr;
+    if (cudaMalloc(&d_ptr, static_cast<size_t>(n) * sizeof(half)) != cudaSuccess) {
+        IMP_LOG_ERROR("Vision: cudaMalloc failed for transposed %zu bytes",
+                      static_cast<size_t>(n) * sizeof(half));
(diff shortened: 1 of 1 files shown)


kekzl:
## Follow-up commit (69e1a91): orientation-aware projector + gemma4v guard

While verifying Gemma-4 vision, I found **Gemma-4 (`projector_type=gemma4v`) is a structurally different vision encoder** — not just renamed tensors:

| | gemma3 (this PR fixes) | gemma4v |
|---|---|---|
| image / patch | 896 / 14 → 4096 patches | 224 / 16 → 196 patches |
| normalization | mean/std 0.5 | mean 0 / std 1 |
| projector orient. | `[out, in]` (transposed) | `[in, out]` (standard) |
| pre-proj norm | `mm.soft_emb_norm` (RMSNorm) | `v.std_scale`/`v.std_bias` |
| post_ln | yes | none |
| pos embed | `[1152, 4096]` 2D | `[1152, 10240, 2]` 3D |

So two hardening changes:

1. **Orientation-aware transpose** — the projector transpose now keys on which dim equals the SigLIP hidden size, instead of unconditionally transposing. The original unconditional transpose would have *corrupted* standard-orientation projectors (gemma4v, LLaVA).
2. **Reject `gemma4v` at load** with a clear message instead of silently hallucinating. Full gemma4v support is a separate, larger feature (different pooling, 3D positional embeddings, std-scale standardization).

Verified: gemma-3-4b still describes images correctly (regression-clean); gemma-4-26B-A4B mmproj now fails cleanly with *"not supported yet — use a gemma3 mmproj"*.

kekzl:
## Follow-up commit (fc6db45): full Gemma-4 (gemma4v) vision support

The earlier commit *rejected* gemma4v as unsupported. This commit instead **implements it fully** — a complete second vision encoder path. (Squash-merge collapses the reject→support churn.)

gemma4v is structurally different from gemma3's SigLIP:
- **RMSNorm** blocks (not LayerNorm), **per-head q/k/v RMSNorm**, **2D axial NEOX RoPE** (θ=100), Gemma **sandwich post-norms**, **GeGLU** FFN, **scale-1** attention.
- Axial learned position tables, avg-pool(3) → ×√D → `(x-std_bias)·std_scale` → pre-proj RMSNorm → linear projector.
- 768px square preprocessing → 48×48 grid → 256 image tokens.

New `encode_impl_gemma4v` + kernels (per-head/weightless RMSNorm, 2D RoPE, axial-pos-add, GeGLU multiply, **fused FP32 projector tail**).

**Key bug found & fixed (systematic-debugging):** output was `<pad>` because the encoder produced **7 fully-NaN tokens**. Localized via stage instrumentation: Gemma vision activations reach absmax ~3000, and the ×√D pooler scale overflows FP16 (→ inf → NaN through RMSNorm's `inf×0`). Computing the projector tail in FP32 (fused kernel) fixes it — NaN 19712→0.

Injection: resolve Gemma-4 image tokens (`<|image>` begin / `<|image|>` repeated soft / `<image|>` end) in chat_template + vision_pipeline.

**Verified** (gemma-4-26B-A4B Q4_K_M + mmproj, greedy): cat → *"brown and black tabby cat"*, pizza → *"tomato sauce, melted cheese, basil"*. gemma3 regression-clean, `make verify-fast` OK. Spec: `docs/vision_gemma4v_spec.md`.

PR #494, opened 2026-05-31
fix(nvfp4): Phi-4 fused-projection scale split + chat generation-prompt

kekzl:
## Fix Phi-4-NVFP4 (Phi3ForCausalLM): fused-projection scale split + chat generation prompt

Phi-4-reasoning-plus-NVFP4 was listed as broken (degenerate, prompt-blind output). Root-caused two independent bugs.

### Bug 1 — fused qkv/gate_up `weight_scale` never split to sub-projections (`53767eaf`)
Phi-3/Phi-4 fuse q/k/v into `qkv_proj` and gate/up into `gate_up_proj` with a **single** per-group `weight_scale` (`[Q+K+V, K/16]` / `[2·ff, K/16]`). The loader routed that scale only to the primary slot (`wq` / `w_gate`), so the split targets **wk, wv, w_up** had no `weight_scale` in `nvfp4_scratch_` → Phase 0 skipped promoting them to NVFP4 → garbage K/V and up-projection.

**Fix:** slice the fused `weight_scale` by output-row range and route each slice to its slot.

**Evidence:**
- Promoted NVFP4 weights **160 → 280** (was 4/layer, now the full 7/layer × 40).
- Slice geometry verified against checkpoint shapes (`weight_scale` `[7680, 320]` / `[35840, 320]`).
- **Completion-mode output now correct + prompt-sensitive**: `2+2`→`4`, `largest city in Germany`→`Berlin`, `one, two, three, four,`→`five, six, seven…`.
- Only affects fused-projection models (Phi-3/Phi-4); separate-q/k/v models (Qwen3 etc.) unaffected (regression-checked: Qwen3-8B still coherent).

### Bug 2 — `{% generation %}` tags broke `add_generation_prompt` (`3c649650`)
Phi-4's chat template wraps the assistant span in `{% generation %}…{% endgeneration %}` (a HF loss-mask marker, render-neutral). imp's minimal Jinja2 engine doesn't know the tag; the unknown opener derailed block nesting so the trailing `{% if add_generation_prompt %}{{ '<|im_start|>assistant<|im_sep|>' }}{% endif %}` was dropped → no generation prompt → model emitted role markers as text.

**Fix:** strip the render-neutral `{% generation %}`/`{% endgeneration %}` tags before parsing.

**Evidence:** rendered chat prompt now correctly ends with `<|im_start|>assistant<|im_sep|>` (previously ended at `<|im_end|>`).

### Honest residual (not blocking)
With both fixes the model **loads correctly, is promoted fully, and is coherent + prompt-sensitive in completion mode**, and the chat prompt is now structurally correct. However, Phi-4-reasoning-plus **chat-mode** output on trivial prompts (e.g. "15+27") is still meta-heavy/poor under its mandatory reasoning system prompt — this appears inherent to the reasoning model (designed for long `<think>` on hard problems) and/or a deeper decoding nuance, independent of the two bugs fixed here. Completion-mode correctness proves the weights/load path are fixed.

Build green (CUDA 13.3). No GPU CI runner — validation above is local.

### DIFF
--- a/src/model/chat_template.cpp
+++ b/src/model/chat_template.cpp
@@ -125,6 +125,36 @@ ChatTemplateFamily ChatTemplate::parse_family(const std::string& name) {
     return ChatTemplateFamily::RAW;
 }
 
+// imp's minimal Jinja2 engine has no {% generation %}/{% endgeneration %} block
+// tags — a HuggingFace chat-template extension that marks the assistant-response
+// span for training-time loss masking (render-neutral). Phi-4-reasoning places
+// them inside the assistant branch; the unknown opening tag derails block nesting
+// so the trailing {% if add_generation_prompt %} is dropped → the assistant
+// generation prompt is never appended → the model emits role markers as text.
+// Strip the (render-neutral) tags before parsing.
+static std::string strip_generation_tags(const std::string& s) {
+    std::string out;
+    out.reserve(s.size());
+    size_t i = 0;
+    while (i < s.size()) {
+        if (s[i] == '{' && i + 1 < s.size() && s[i + 1] == '%') {
+            size_t close = s.find("%}", i + 2);
+            if (close != std::string::npos) {
+                std::string inner = s.substr(i + 2, close - (i + 2));
+                size_t a = inner.find_first_not_of(" \t\r\n-");
+                size_t b = inner.find_last_not_of(" \t\r\n-");
+                std::string kw = (a == std::string::npos) ? std::string() : inner.substr(a, b - a + 1);
+                if (kw == "generation" || kw == "endgeneration") {
+                    i = close + 2;  // drop the whole tag, render its inner body as-is
+                    continue;
+                }
+            }
+        }
+        out.push_back(s[i++]);
+    }
+    return out;
+}
+
 bool ChatTemplate::init(ChatTemplateFamily family, const Tokenizer& tokenizer, const std::string& jinja_str) {
     family_ = family;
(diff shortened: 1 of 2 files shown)


kekzl:
---
### Correction to the residual (investigated further)
My earlier note that the chat residual was "model-inherent reasoning behavior" was **wrong**. Further root-cause investigation (context-length sweep with a stable factual probe, `--chat-template none`) shows a **third, distinct, Phi-4-specific bug**: attention degrades beyond ~128–256 tokens of context.

- Factual probe "…N tokens of varied context… the capital of France is" → **8 tok: "Paris" ✓ · 113 tok: "Paris" ✓ · 269 tok: "the sky…" ✗** (echoes early context, ignores the recent question).
- **Phi-4-specific**: identical 269-token probe on Qwen3-8B → "Paris" ✓.
- **Both attention paths fail** at 269 tok (forced FMHA → "a. The. The"; forced cuBLAS → "the sky") → not path-specific → upstream Q/K/V or position encoding.
- Ruled out: rope_theta (500000, loaded correctly), partial_rotary_factor (1.0), head_dim (infers 128 correctly).

This is consistent with a subtle K/V-projection or position error that's tolerable at short context but compounds at length. It is **separate from the two bugs fixed in this PR** (which are correct and necessary — short-context completion is fully coherent + prompt-sensitive). Root-causing it needs a numerical K/V comparison against a reference (transformers) — tracked as follow-up, not in this PR.

PR #506, opened 2026-06-04
fix(server): model-reload SIGSEGV + VRAM retention; strict OpenAI model semantics (no auto-swap)

kekzl:
## Problem

Requesting a different model on the server (`[auto-swap]`) killed the process: SIGSEGV (exit 139) with no error output, docker restart-policy relaunched it — clients saw connection drops. Even when it didn't crash, swaps failed with "Failed to upload token embedding" although the new model fit easily.

## Root causes (reproduced under gdb)

1. **Dangling cuBLAS stream → SIGSEGV.** The process-global attention-cuBLAS handle keeps the stream of the destroyed engine. The next engine's `attention_cublas_prewarm()` runs its dummy GemmBatchedEx on that stream → `cuStreamGetGreenCtx` segfaults in libcuda. Only triggers after ≥1 inference (the handle is stream-bound lazily) — which is why cold swap tests never caught it.
   ```
   #2  cuStreamGetGreenCtx
   #6  cublasLtHSHMatmulAlgoGetHeuristicForStream
   #14 imp::attention_cublas_prewarm()
   #15 imp::Engine::init(...)
   ```
2. **Default mempool retention → reload starvation.** Weights are allocated via `cudaMallocAsync` from the default pool (release threshold UINT64_MAX). Freeing model+context parked **~28 GB** in the pool; `cudaMemGetInfo`-based sizing and the upload oversubscription gate then saw 1.5 GB free on a 32 GB card. Fix: `cudaMemPoolTrimTo(pool, 0)` in `imp_context_free`/`imp_model_free`.

## API change (breaking, intentional)

Auto-swap removed — inference requests never touch model lifecycle (the old behavior cancelled every in-flight request on swap):
- `/v1/models` exposes **only the loaded model** (was: full models_dir listing)
- requesting another model → **404 `model_not_found`** (OpenAI error shape); the silent llama.cpp-style serve-with-current fallback is gone
- auto-load only when no model is loaded; switching models = restart with another `--model`

## Verification

- New `EngineRelaunchTest`: load → generate → free → **assert VRAM returned** → reload → generate. Segfaulted/failed before, green now.
- Live server matrix: `/v1/models` single entry · correct model 200 · foreign model 404 · **concurrent stream + foreign-model request: stream completes (`stop` + `[DONE]`), server alive, RestartCount=0**
- `make verify-fast` OK (no hot-path changes; decode/prefill untouched)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -537,6 +537,7 @@ if(IMP_BUILD_TESTS)
         tests/test_chunked_prefill.cu
         tests/test_mtp_forward.cpp
         tests/test_api_generate.cpp
+        tests/test_engine_relaunch.cpp
     )
 
     set(IMP_TEST_BINARIES
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -194,8 +194,11 @@ Endpoints: `/v1/chat/completions`, `/v1/completions`, `/v1/embeddings`,
 non-streaming), `/tokenize`, `/detokenize`, `/health`. Tool/function
 calling, streaming usage stats, logprobs, and API-key auth
 (`--api-key`) supported.
-`/v1/models` lists available GGUF and SafeTensors models in the models
-directory.
+`/v1/models` lists the model the server is serving (OpenAI semantics: the
+server exposes exactly what it can serve). Requests must name that model —
+any other `model` value gets `404 model_not_found`; inference requests never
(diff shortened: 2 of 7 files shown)


kekzl:
Superseded by rebased branch (force-push restricted) — see follow-up PR.

Issue #547, opened 2026-06-06
gpt-oss-20b (GptOssForCausalLM) unsupported — GOAL.md hero model needs arch + SafeTensors-MXFP4 MoE path


kekzl:
## What

https://huggingface.co/openai/gpt-oss-20b does not load: `model_type: gpt_oss` / `architectures: ["GptOssForCausalLM"]` has no entry in `src/model/model_arch.h` (enum ends at LLAMA4/GENERIC). The only existing code touchpoint is the MXFP4 quantization-config *detection* in `hf_config_loader` (`src/model/hf_config_loader.cpp:944`, audit gap #13) — metadata parsing only, no compute path behind it.

`GOAL.md` lists gpt-oss-20b as a hero model with the rationale **"First-class MXFP4 path"**, and "If a hero model regresses against any competitor on the primary metric, that is a release blocker" — so this is a tracked gap, not a nice-to-have.

## Why it's attractive on the 5090

- 21B total / ~3.6B active (MoE 32 experts, top-4) — checkpoint ~13 GB, fits 32 GB with huge ctx headroom.
- MoE expert weights ship **natively MXFP4** (block/scales uint8 tensors in SafeTensors) — maps directly onto the sm_120a `mma.sync mxf4nvf4` story; no community-quant quality lottery (unlike the GGUF-MXFP4 family, #qwen35-mxfp4).
- Strong agent/tool-use model — fits the agent-backend track (#522).

## Architecture requirements (vs. what imp has today)

| Feature | gpt-oss-20b | imp status |
|---|---|---|
| MoE 32 experts, top-4, softmax-after-topk | router with bias | MoE infra exists (Qwen3-MoE etc.) — wiring needed |
| **MXFP4 expert weights in SafeTensors** (`*.weight_blocks` + `*.weight_scales`) | core format | loader detects the quant config but has no dequant/compute path; existing MXFP4 kernels are GGUF-side and quality-flagged |
| **Alternating attention**: sliding-window (128) on even layers, full on odd | per-layer pattern | per-layer sliding-window pattern not in ModelConfig |
| **Attention sinks** (learned per-head sink logit in softmax) | required for quality | not implemented in any attention path (FA2/fp8-FMHA/decode) |
| GQA 64Q/8KV, head_dim 64, **attention bias** | — | hd=64 paths exist; bias support to verify |
| RoPE + **YaRN** to 131k ctx | — | YaRN support to verify for SafeTensors path |
[...]

PR #572, opened 2026-06-06
feat(gpt-oss): gpt-oss-20b support — MXFP4 experts, sinks, Harmony, YaRN/splitk/FP16-range fixes (#547)

kekzl:
Implements gpt-oss-20b (SafeTensors MXFP4) support — issue #547.

## Architecture (GPT_OSS, arch id 14)
- HF config mapping (GptOssForCausalLM): YaRN(32x), per-layer SWA-128 (even layers), 64q/8kv heads @ hd=64, 32-expert top-4 MoE, attention sinks, q/k/v/o biases, untied lm_head.
- Weight map: `self_attn.sinks`, `mlp.router.{weight,bias}`, packed MXFP4 expert `gate_up/down_proj_{blocks,scales,bias}`, o_bias. 459/459 tensors assigned; fused gate_up bias de-interleaved at load.

## MXFP4 -> NVFP4 expert conversion (`src/quant/gpt_oss_mxfp4_convert.cu`)
Nibbles are bit-identical (both e2m1 linear pair order); only scales convert: 1 ue8m0 -> 2 e4m3 micro-scales + per-expert FP32 tensor scale `2^(max_u-127-8)` (exact powers of two, zero loss; clamps logged, none fire on the 20b checkpoint). gate_up rows are interleaved (g0,u0,g1,...) and de-interleaved during conversion. Both MoE paths consume the result: batch-prefill dequant GEMM + decode GEMVs.

## Forward specifics
- **Router**: true linear bias on the logits (`moe_add_logit_bias`) *before* top-k — affects selection AND weights (deliberately distinct from DeepSeek's selection-only `score_bias`); registry `expert_weights_norm=true` makes imp's renormalized top-k softmax exactly HF's topk->softmax.
- **Clamped GLU** (`gpt_oss_glu`): `gate<=7`, `up in [-7,7]`, `(up+1) * gate * sigmoid(1.702*gate)`.
- **Per-expert biases**: indexed kernel (decode, via expert_indices) + sorted kernel (prefill, binary search over expert_offsets).
- **Attention sinks**: virtual extra softmax column (denominator += `exp(sink-max)`, column dropped) in the cuBLAS prefill softmax kernels and the paged decode kernels (GQA, cluster via `crosswarp_reduce_and_write`, split-K reduce). Prefill routing forces cuBLAS when sinks are present.
- **o_bias** after Wo+residual (same seam as the LoRA O hook).

## Bugs found & fixed (all latent, gpt-oss was the first model to hit them)
1. **HF `rope_freq_scale` convention inverted** (`hf_config_loader`): imp stores the FACTOR (GGUF convention, kernel applies 1/factor); the HF loader stored 1/factor -> YaRN interpolated dims rotated factor^2=1024x too fast, mscale 0.653 instead of 1.347. This was the dominant quality bug: PPL 104 -> 7.7 (HF BF16 reference: 6.71 on the same corpus).
2. **FP16 residual overflow**: gpt-oss's massive activations reach |h| > 65504 by L23 (hidden L2 trace matches HF: ~42k jump at L7) -> inf/NaN logits. Fix: exact 2^-4 residual rescale (embeddings via `embed_scale`, Wo/o_bias/down_bias host-scaled, expert-down via converter tensor_scales). FP16 power-of-two scaling is an exponent shift (lossless); RMSNorm scale-invariance keeps the model output mathematically identical — verified by unchanged PPL, minus the inf/NaN class.
3. **Split-K paged attention hd=64 misaligned cp.async**: `cp_async_ca_8` from `lane_offset = lane*2 halves` (4-byte aligned) raises `cudaErrorMisalignedAddress` for ELEMS==2. New `cp_async_ca_4` + ELEMS-gated selection. Whole zoo is hd>=128, so this never fired before. New unit tests `PagedAttentionTest.GQA_SplitK_HD64{,_Sinks}` (split-K was previously never activated in tests — no scratch).
4. **Fused rope-KV decode path ignores YaRN**: `rope_q_only_fp16_kernel`/`write_kv_cache_rope_fused_kernel` know only theta+scale -> decode-written K was plain-RoPE while prefill K was YaRN-correct -> degeneration from token 2. YaRN models now take the separate full-rope path.
5. **`swa_layers` was honored only for GEMMA4** -> gpt-oss would have gotten SWA-128 on all 24 layers instead of even ones.
6. **`force_cublas_decode` debug arm broke SWA** (causal=false/q_offset=0 never triggers the window mask) — now causal=true @ q_offset=ctx-1 + sinks pass-through, keeping it a faithful reference.

## Harmony chat template
New `HARMONY` family ( `<|start|>role<|message|>...<|end|>`, system turn with channel contract, user `system` -> developer role, assistant history as final channel). Detected via `<|channel|>` before the Phi `<|end|>` check. Stops: `<|return|>` + `<|call|>` — `<|end|>` deliberately NOT a stop (separates analysis from final). The banned-token list now keeps the Harmony structure tokens (the model is trained to emit them; banning trapped generation in an endless analysis channel). imp's mini-Jinja renders the shipped `chat_template.jinja` correctly as well.

## Validation
- Teacher-forced PPL 7.67 vs HF transformers BF16-dequant CPU reference 6.71 (same corpus, same tokens; engine-class delta).
- Greedy + sampled chat produce clean Harmony output: `<|channel|>analysis<|message|>...<|end|><|start|>assistant<|channel|>final<|message|>The capital of France is **Paris**.`
- 220-token decode crosses the SWA-128 boundary coherently, graphs ON == OFF behavior.
- Regression: Qwen3-8B Q8 tg 275 tok/s (baseline 268), pp 8437; "17+25 -> 42"; gemma-3-12b SWA "Paris" — all clean. `verify-fast` gtest filter PASS.

## Follow-up (tracked in #547)
Server-side Harmony channel split (analysis -> `reasoning_content`, final -> `content`) — the existing think-machinery is `<think>`-token based; gpt-oss needs the `<|end|>`-separator variant + header stripping.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -270,6 +270,7 @@ set(IMP_EXEC_SOURCES
     src/exec/executor_perplexity.cu
     src/exec/executor_lora.cu
     src/lora/lora_adapter.cpp
+    src/quant/gpt_oss_mxfp4_convert.cu
     src/exec/executor_ffn.cu
     src/exec/executor_ssm_gdn.cu
     src/exec/executor_forward_moe.cu
--- a/include/imp/types.h
+++ b/include/imp/types.h
@@ -37,6 +37,7 @@ typedef enum {
     IMP_ARCH_QWEN35_MOE = 11,
     IMP_ARCH_GEMMA4 = 12,
     IMP_ARCH_QWEN36_MOE = 13,
+    IMP_ARCH_GPT_OSS = 14,
 } ImpModelArch;
 
 typedef enum {
--- a/src/compute/activation.cu
+++ b/src/compute/activation.cu
@@ -138,6 +138,54 @@ __global__ void geglu_fp32_vec4_kernel(const float* __restrict__ gate, const flo
 }
 
 // --------------------------------------------------------------------------
+// gpt-oss clamped GLU kernels (issue #547):
+//   gate_c = min(gate, 7);  up_c = clamp(up, -7, 7)
+//   out = (up_c + 1) * gate_c * sigmoid(1.702 * gate_c)
+// --------------------------------------------------------------------------
+__device__ __forceinline__ float gpt_oss_glu_elem(float g, float u) {
+    constexpr float kLimit = 7.0f;
+    constexpr float kAlpha = 1.702f;
+    g = fminf(g, kLimit);
+    u = fminf(fmaxf(u, -kLimit), kLimit);
+    float glu = g / (1.0f + __expf(-kAlpha * g));
+    return (u + 1.0f) * glu;
+}
+
+__global__ void gpt_oss_glu_fp16_kernel(const __half* __restrict__ gate, const __half* __restrict__ up,
+                                        __half* __restrict__ out, int64_t n) {
+    const int64_t idx = (static_cast<int64_t>(blockIdx.x) * blockDim.x + threadIdx.x) * 2;
+    if (idx + 1 < n) {
+        float2 gf = __half22float2(*reinterpret_cast<const __half2*>(gate + idx));
+        float2 uf = __half22float2(*reinterpret_cast<const __half2*>(up + idx));
+        *reinterpret_cast<__half2*>(out + idx) = __float22half2_rn(
+            make_float2(gpt_oss_glu_elem(gf.x, uf.x), gpt_oss_glu_elem(gf.y, uf.y)));
+    } else if (idx < n) {
+        out[idx] = __float2half(gpt_oss_glu_elem(__half2float(gate[idx]), __half2float(up[idx])));
+    }
+}
+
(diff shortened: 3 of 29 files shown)


kekzl:
Server-Harmony-Split nachgeschoben (gleicher PR): `split_harmony_segments` (non-streaming + completions) + Token-Statemachine in beiden Streaming-Pfaden (OpenAI SSE via `emit_reasoning`, Anthropic via `emit_thinking`). E2E verifiziert: non-stream `content='The capital of France is **Paris**.'` / `reasoning_content='The user asks…'`, stream sauber getrennt, Multi-Turn ok ('Your name is Otto.'). Damit ist das Follow-up aus der PR-Beschreibung bereits enthalten.

PR #621, opened 2026-06-07
VRAM cache rebuild: Stage 1 (one tier truth) + Stage 2 (RAII ownership)

kekzl:
## Why

Audit of the weight-cache / VRAM layer (`docs/audit/vram_cache_structure_2026_06_07.md`, verified against code + real bench logs) found the cache layer mid-refactor with the migration stalled:

- **G1** — three diverging tier oracles: `StoragePlanner` (source-aware, correct, but "diagnostic only"), the source-blind `vram_budget.cpp` heuristic (actually allocates), and a *second* `nvfp4_beneficial` copy. Scattered `if(arch==…)` rules across `engine_init_resolver` + the phase builders — the exact class that caused the rejected FP16-gate diff this morning (north-star decode −15.5% + gemma-3 crash) and the historical #428×#434 regression.
- **G2** — Phase 4b's drop-source diagnostic claimed "7.6 GB could be freed (overlay covers prefill + decode)" — false since IMMA raw-read prefill (#617) reads the GGUF source directly.
- **G3** — dead CUTLASS scale-factor buffers on GGUF (Q8: 253 tensors / 451 MiB never read).

Design: `docs/superpowers/specs/2026-06-07-vram-cache-architecture-design.md`. Strictly quality-neutral, perf-gated, staged. This PR is **Stage 1** — make the plan authoritative for the overlay-tier decision, starting with the FP16 phase, fold the scattered arch rules into one pass, and drop the dead posts that fall out as a consequence.

## What

- **`StoragePlan` becomes the held source of truth** (1.1/1.2): built once after Phase 0 (so prequant-NVFP4 weights carry `QType::NVFP4`), with O(1) `tier_of(src)` lookups and a per-tensor plan/actual **parity diagnostic** that proved the migration map before any builder was switched.
- **`apply_arch_rules_`** (1.2): one pass folding the scattered overlay rules into the plan — FP8-unavailable → FP16 fallback, `nvfp4_lm_head` → LM_HEAD NVFP4 (GDN-gated), gemma-3 NVFP4-from-FP16 companion. One function decides; builders read.
- **Phase 1 FP16 cache reads the plan** (1.3a): three inline rules (`nvfp4_beneficial` early-return, the GEMMA3 exception, the FP8-widening) collapse to one plan query.
- **G3: skip dead CUTLASS SF** (1.4): Q8_0 + `q8_imma` weights do M>1 prefill via IMMA raw-read, so their CUTLASS SF buffer is never read — Phase 3b skips it (keyed off the plan's `source_qtype`). CUTLASS stays for native-NVFP4 and Q6_K/Q5_K (their actual prefill path).
- **G2: honest diagnostics** (1.5): renamed to "decode-redundancy … kept resident for M>1 prefill (upper bound, not freeable)"; Phase 4b comment corrected to match the code.

## Verification (GPU, healthy host: mem 13801, 5090)

| Model | Parity | Coherence | Decode (vs baseline) | Prefill |
|---|---|---|---|---|
| Qwen3-8B Q8_0 | mismatch=0 | ✓ | 265/267 (~262) | 12.1k (~12k) |
| native-NVFP4 (cortecs) | mismatch=0 | ✓ | — | — |
| Qwen3-14B Q6_K (north-star) | mismatch=0 | ✓ | 143 (~145) | 6.5k (~6.5k) |
| gemma-3-12b Q4_K_M | mismatch=1* | ✓ "Paris.<end>" | — | — |

\*the one remainder is `TOK_EMBED` (tied with lm_head→NVFP4), a benign 1-tensor diagnostic-only edge case.

- **G3 win:** Qwen3-8B Q8_0 `cutlass_nvfp4` 253 → 0 tensors (451 MiB freed), prefill/decode neutral (proves the SF was dead). Qwen3-14B Q6_K unaffected.
- No IMA/NaN/CUDA errors. **`make verify-fast`: OK** (gtest + perf gate + degeneration smoke).

## Out of scope (follow-up PRs, per the staged design)

- Phase 3 (NVFP4) plan-driven — needs a plan-schema extension (a single `tier` can't express "prefill via CUTLASS, decode via NVFP4" for native-NVFP4).
- Phase 2 (FP8, no-op on sm_120), `vram_budget` sizing on plan bytes, Stage 2 (RAII arena ownership), Stage 3 (honest budget).

### DIFF
--- a/docs/audit/vram_cache_structure_2026_06_07.md
+++ b/docs/audit/vram_cache_structure_2026_06_07.md
@@ -1,0 +1,135 @@
+# VRAM / weight-cache structure audit — 2026-06-07
+
+Scope: where model-weight VRAM goes, which copies coexist, and where the
+caching/budget architecture has lost structure. All findings verified against
+source AND real bench logs (Qwen3-8B Q8_0 GGUF, Qwen3-14B Q6_K GGUF on clean
+`main`). Diagnostic-only; no code changed.
+
+## TL;DR
+
+Per GGUF weight, up to **three** representations are resident at once:
+the original GGUF blocks (used by prefill), an NVFP4 decode cache (used by
+decode), and a CUTLASS scale-factor buffer (**used by nobody on GGUF**). On top
+of that, three independent systems decide tiers and budget, only the weakest of
+which actually drives allocation, and two cleanup passes that look load-bearing
+are dead no-ops. The owner's instinct ("null Struktur") is correct: the cache
+layer is mid-refactor and the migration stalled.
+
+Measured residency (clean main, real logs):
+
+| Model | GGUF source (prefill) | NVFP4 decode | CUTLASS SF | weight held ≈ |
+|-------|-----------------------|--------------|------------|---------------|
+| Qwen3-8B Q8_0  | ~8.0 GB | 4.06 GB | 0.45 GB (dead) | **1.55×** |
+| Qwen3-14B Q6_K | ~9.76 GB | 6.69 GB | 0.83 GB (dead) | **1.8×** |
+
+`GPU memory: 16396 MiB used` for an 8B-Q8 model that weighs 8.3 GB on disk.
+
+---
+
+## Gap G1 — three tier-decision systems, only the worst one rules
+
+There are three places that decide "which tier does weight X get", and they
+disagree by design:
+
+1. **`StoragePlanner`** (`src/runtime/storage_planner.cpp`) — fully implemented,
+   source-qtype-aware (`effective_capabilities(kind, source_qtype)`), has a
+   budget-pressure downgrade loop. **Completely unused.**
+   `src/runtime/vram_budget.cpp:174`: *"Heuristic still drives allocation;
+   planner output diagnostic only (5.1.5)."*
+2. **Heuristic** (`vram_budget.cpp:106-271`) — source-*blind*
+   (`nvfp4_beneficial(qtype)`, never consults `kind` or the capability table).
+   This is what actually sizes the caches.
+3. **Runtime `nvfp4_beneficial`** — a **second copy** of the same predicate at
+   `src/exec/pre_dequant_internal.h:98`, called per-tensor during the Phase 0-3
+   build to decide which weights physically get an NVFP4 slot. Identical logic to
+   the `vram_budget.cpp:108` lambda, different file → silent drift risk.
+
+Live divergence in the logs: planner projects **10055 MiB**, heuristic projects
+**7504 + 833 MiB**. Nobody reconciles the gap; the planner's answer is logged and
+thrown away. The planner has been "diagnostic only" since commit 5.1.5 and the
+flip to make it authoritative ("retire heuristic") was never started.
+
+## Gap G2 — Phase 4b "drop redundant GGUF source" is a dead no-op
+
+`pre_dequant_phase4b_drop_redundant_sources_` (`pre_dequant_phase4_tensor_registry.cu:463`)
+is wired into the pipeline with `actually_free = true`, and a sibling diagnostic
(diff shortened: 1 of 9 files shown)


kekzl:
Added **Stage 1.6 — Phase 3 NVFP4 collect reads the plan** (commit 18dc5102).

Second builder migrated to plan-driven, same methodology: `collect_weight_nvfp4`'s inline `nvfp4_beneficial` + `force_beneficial` bypass → one plan query. The `nvfp4_ssm_proj` opt-in is now encoded in `apply_arch_rules_`; `nvfp4_attn_proj` is a separate path, untouched.

Verified (parity == legacy, cache sizes identical → decode-perf identical):
- Q8 253 / native-NVFP4 253 / Q6_K 105 / gemma-3 49 tensors, mismatch≤1
- **qwen3.6-35B GGUF hybrid + `nvfp4_ssm_proj=ON`: 281 tensors (ssm cached), parity mismatch=0, coherent** — the +53% hybrid-decode opt-in is intact.
- All coherent, no IMA/NaN, `verify-fast` OK.

Both main overlay-tier builders (FP16 + NVFP4) now read the plan.

kekzl:
## Stage 2 — RAII arena ownership (commits dc8437e7…8d77832b)

All **8 weight caches** now own their storage via a new move-only `VramOwned` (`src/memory/vram_owned.h`). Design: `docs/superpowers/specs/2026-06-08-vram-cache-stage2-raii-ownership-design.md`.

| Stage | Cache |
|---|---|
| foundation | VramOwned + drop dead q4k_imma map (never populated) |
| 2.1 | cutlass_nvfp4 (borrowed-SF, MoE re-key) |
| 2.2 | nvfp4 — central decode cache (9 hot-path copies → `borrow_view()`) |
| 2.3 | cutlass_mxfp4 + nvfp4_moe |
| 2.4 | fp16 (+bulk) + fused_kv/gate_up (parallel keyed-owner; registry transfer via `release()`) |
| 2.5 | fp8 (+bulks) |

**Outcome:** the `borrowed`/`owned`/`sf_borrowed` bool-flag geflecht is gone (ownership is the type); the ~150-line `free_buffers` teardown + every bulk range-check collapse to `clear()`/`reset()`; **double-free is now a compile error** (move-only surfaced ~12 latent hot-path copies); the **#507 model-swap leak class is structurally closed**.

**Verified per stage:** parity mismatch≤1, coherent across dense/MoE/GGUF/gemma-3, CLEAN (no IMA/NaN/double-free), `EngineRelaunchTest.ReloadAfterInferenceReleasesVramAndDoesNotCrash` PASSED, `verify-fast` OK.

Out of scope (own follow-up): Stage 3 — make the planner downgrade loop drive real allocation + replace the `reserve` fudge.

kekzl:
## Stage 3 — G1 closed (commit 1d4d1bf8)

Consolidated the last duplicate: `nvfp4_beneficial` existed in **three** places (pre_dequant_internal.h function, vram_budget.cpp lambda, engine_init_resolver.cpp inline expression) → one definition in `core/qtype.h`.

**Investigated + deliberately did NOT do** "planner drives the budget": the planner's `projected_vram_bytes` (ideal overlay upper bound — e.g. ~21 GiB on gemma-3-12B as all Q4_K→FP16) and the heuristic's `nvfp4_estimate` (NVFP4 cache + KV sizing) measure **different things**. Driving the budget off the planner total would crush KV to zero → OOM. The planner owns the **tier decision** (Stage 1); the heuristic owns cache **sizing** — two jobs, not three oracles. Corrected the misleading "diagnostic only (5.1.5)" log accordingly.

Parity unchanged, CLEAN, verify-fast OK — zero behaviour change.

---
**This PR is now the complete VRAM-cache rebuild (G1–G5):** one tier truth (Stage 1) · RAII ownership across all 8 caches (Stage 2) · single policy + honest budget (Stage 3).

PR #1095, opened 2026-07-26
feat(constrain): GBNF grammar-constrained decoding (response_format grammar / guided_grammar)

kekzl:
Closes roadmap gap 8 — the last constrained-decoding surface imp was missing.

## Why this needed a new engine

A regex covers the formats agents pin most often (IDs, enums, dates, diff headers), and #1091 shipped that. But a regex is *regular by definition*: a nested expression language, a bracket-balanced DSL, any recursive format needs a **stack**, which `RegexNfa` structurally cannot have. So unlike the regex constrainer — which was a decode-time wrapper over an engine already in the tree — this one is a real second engine: a nondeterministic pushdown simulator behind the same `apply_mask` contract as the JSON, schema and regex constrainers.

## Surface

Three spellings, all accepted, so a client written against llama.cpp or vLLM works unchanged:

```jsonc
{"response_format": {"type": "grammar", "grammar": "root ::= ..."}}   // imp
{"grammar": "root ::= ..."}                                          // llama.cpp
{"guided_grammar": "root ::= ..."}                                   // vLLM
```

GBNF subset: alternation, string literals, character classes (`[a-z]`, `[^"]`), `.`, rule refs, grouping, `*` `+` `?` `{m}` `{m,}` `{m,n}`, comments, `\xNN`/`\uNNNN` escapes.

## What actually took the time

**Cost.** The first correct cut spent **333 ms building ONE token mask** inside a JSON string — a cold mask simulates the whole 151k-token vocabulary, and the naive vector-of-frames stack allocated a few dozen times per byte. It did not look like a slow feature, it looked like a hung request. Two changes fixed it:

- parse stacks are **hash-consed to integers**, so a step is integer work and the state set is a flat vector that is cheap to copy, compare and cache;
- each stack's **successor set is memoised** — a transition's target does not depend on *which* character took it, only on whether it matched.

Same mask, same grammar: **333 ms → 12 ms** (27×). Measured with a synthetic 151k-token vocabulary; harness kept out of the tree, numbers in the commit body.

**Refusal discipline.** A grammar the simulator cannot expand must not become a hung request or a silently different language, so these are rejected at compile time and the request then decodes *unconstrained* rather than under a wrong grammar: left recursion (direct, indirect through a nullable prefix, and the star-over-nullable spelling), undefined rules, missing `root`, grammar-bomb repetition bounds.

**Correctness at the byte level.** EOS is gated on a complete derivation, so generation cannot stop half-way through the format. UTF-8 is assembled across token boundaries (a BPE token can end mid-character) and **overlong encodings are rejected** — otherwise a token could spell a forbidden character in a longer form and walk straight past the mask.

## Bypass paths

Per the `server-api` checklist, since a constrainer is only as good as the paths that cannot route around it: spec-ngram router, graph-loop router, the constrained-pipeline and scheduler `InferenceState` sites, `ensure_constraints_` / `pipeline_row_eligible_`, the thinking default (structured output must suppress it, or a reasoning model spends the budget thinking and the caller gets prose), and pooled-manager state.

The four copies of the constrainer chain in `executor.cu` are now **one `apply_constraint_mask` helper**. A new constrainer previously had to be added in four sampling paths and two of them are easy to miss — that is exactly how an unmasked path ships. Net effect on that file: −10 code LOC, back under the size gate.

## Verification

- **22 CPU-lane tests** in `test-core` (parser, pushdown simulator, UTF-8 assembly, every refusal). CPU-only on purpose — every past constrained-decode bug in this tree escaped CI because its test needed a GPU.
- **Mutation-validated**: 11 deliberate breakages, all caught. One of them found a **real bug the tests had missed** — a recompile cleared the stack arena but not its memoised transitions, so a second grammar on a pooled manager would have been decoded with the first one's. Fixed, and now covered by `RecompilingReplacesTheLanguageCompletely`.
- **Live server** (Qwen3-4B Q8_0 / Qwen3-14B Q6_K): recursive s-expressions come back balanced (`((a b) (c d) e)`), `[0-9]{4}-[0-9]{2}-[0-9]{2}` stops exactly at the bound (`1989-11-09`), a broken grammar degrades to unconstrained output instead of 500, streaming enforces the same grammar, non-ASCII grammars decode correctly, **batch>1 holds four different grammars on four concurrent rows**, and on a think-capable model the answer lands in `content` with empty `reasoning_content`.
- `test-core` 539 passed / 0 failed; `degen_suite --only constrained` unchanged vs `main` (see note); `make verify-fast` green — decode 284.77 vs baseline 287.19 (−0.84%, inside the 3% gate). **No baseline refresh**: nothing here touches the hot path unless a grammar is requested.

## Note, not part of this PR

While running the constrained battery I hit a `json_object` failure on Qwen3-4B-Q8_0 — the free-form JSON FSM accepts a trailing comma (`["camera quality",]`). **It reproduces identically on `ghcr.io/kekzl/imp:latest` (main)**, so it is pre-existing and model-specific, not a regression from this change. Worth its own issue.

https://claude.ai/code/session_01AQf5BRBYTMx7Q5QDWNhMqa


kekzl:
CI note: the `File size` job is red, but it is **inherited from `main`**, not introduced here — `tools/imp-server/tool_call.cpp` sits at 847 code LOC against a hard threshold of 800 and is not in the `[allow]` list. The last three CI runs on `main` (3c9c0b8a, 06ef2911, 8302ea3e) fail the same job, with `Build` green.

This PR moves the gate in the right direction: `src/exec/executor.cu` gained the grammar wiring but came out 10 code LOC *smaller* (595 → 585) because the four copies of the constrainer chain collapsed into one helper, and the grammar engine ships as two translation units (`gbnf_grammar.cpp` simulator + `gbnf_parser.cpp`) so neither trips the warn threshold.

I deliberately did not silently allowlist `tool_call.cpp` here — the `[allow]` list requires a reason, and that file's debt deserves its own decision rather than being buried in a feature PR.

PR #1186, opened 2026-08-01
fix(quantize): refuse stacked-expert checkpoints instead of copying them through

kekzl:
`should_quantize` refuses 3-D tensors with *"3-D MoE expert stack — needs the per-expert path"*, and the summary reports how many it skipped. **Neither ever fired on a real checkpoint.**

The rank check sat *behind* a `.weight` name test, and no stacked checkpoint names its experts that way — gpt-oss ships `mlp.experts.gate_up_proj_blocks`, Gemma-4 `experts.gate_up_proj`. They were rejected one branch earlier as `not a .weight tensor`, which prints nothing, counts nothing, and adds no `excluded_modules` entry.

So the experts were **copied through as BF16 while `hf_quant_config.json` announced NVFP4**. The output loads and runs — it has simply not been quantized, which on a MoE checkpoint is most of the bytes. Same failure class as #1159 (a checkpoint that looks like a result), one branch further up.

### Changes

- The rank check moves in front of the name check, so a stack is diagnosed as a stack whatever it is called.
- A pre-pass over every shard **refuses the checkpoint before anything is written**, naming the tensors and the share of the checkpoint they hold.

The share is computed rather than asserted, and that turned out to matter: on `Qwen3.6-35B-A3B-NVFP4` the only stacks are the MTP head's — **6.4%** — so a refusal claiming "most of the checkpoint" would have been false.

```
refusing: 2 tensor(s) store MoE experts as a 3-D stack — 6.4% of this
checkpoint — and imp has no NVFP4 layout to read stacked experts back from.
Quantizing would copy them through unchanged and still label the result
NVFP4, so the output would claim a size and bandwidth win it does not have.
  mtp.layers.0.mlp.experts.down_proj
  mtp.layers.0.mlp.experts.gate_up_proj
```

### Testing

The tool had **no tests at all**, and this is the part of it with a history of shipping broken checkpoints, so the selection rule moved to `tools/imp-quantize/tensor_policy.{h,cpp}` where a test can reach it. `tests/test_quantize_policy.cpp` runs in the CPU lane and covers the roles #1159 established (MLA latent projections, the router, `gate` vs `gate_proj`) alongside the stack cases.

**Mutation-validated** — restoring the name-before-rank order fails `DiagnosesAStackWhateverItIsNamed`; dropping the `"expert"` name filter from the scanner fails `IgnoresThreeDTensorsThatAreNotExperts`. Both mutations were asserted present in the file before measuring.

**Against real checkpoints:** `Qwen3.6-35B-A3B-NVFP4` is refused as above; `Qwen3-VL-4B-Instruct` still dry-runs to 357 tensors quantized, so the vision tower's many 3-D tensors (patch embeddings, position tables) do not trip the scanner.

`make verify-fast` green against a rebuilt image: decode +0.35%.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -98,6 +98,14 @@ there instead of retelling it.
   violation.
 
 ### Fixed
+- **`imp-quantize` wrote a "quantized" checkpoint whose experts were untouched.**
+  The 3-D refusal sat behind a `.weight` name test, and no real stacked
+  checkpoint names its experts that way (`experts.gate_up_proj`,
+  `..._blocks`), so they were copied through as BF16 with no message, no
+  counter and no exclusion entry — while `hf_quant_config.json` announced
+  NVFP4. Such a checkpoint is now refused before anything is written. The
+  selection rule moved to `tools/imp-quantize/tensor_policy.cpp` and is covered
+  by `tests/test_quantize_policy.cpp` in the CPU lane.
 - **`imp-quantize` silently produced a broken MoE checkpoint.** "MoE is left
   unquantized" only ever applied to 3-D stacked tensors; the HF-standard
   per-expert 2-D layout was quantized and produced a model that loaded and then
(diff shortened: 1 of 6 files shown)


kekzl:
Superseded by #1188 — same commit cherry-picked onto current main. This branch went stale against main (CHANGELOG conflict with #1184) and force-push is blocked on this repo, so it could not be updated in place.

PR #1284, opened 2026-08-06
fix(model): decide the A_log transform from values, not from dtype

### LINKED ISSUE #1282: Qwen3.5-4B (tied embeddings + Qwen3_5 arch): loads, reports tying success, then emits uniform logits (PPL == vocab_size)
`Qwen3.5-4B` (BF16, straight from `Qwen/Qwen3.5-4B`) loads without an error and then emits `<|im_start|>` forever. Teacher-forced perplexity is **248320.0000**, which is exactly `vocab_size` — `mean_nll = 12.4225 = ln(248320)`. That number is the signature of perfectly uniform logits: the output projection produces nothing.

Found while trying to build a BF16 hybrid checkpoint to validate the #1273 quantizer fix against.

## Repro

```
imp-cli --model Qwen3.5-4B --prompt "The capital of France is" --max-tokens 12 --temperature 0
  -> [tok=248045 '<|im_start|>'] x12
imp-cli --model Qwen3.5-4B --perplexity tools/analysis/ppl_corpus_45k.txt
  -> perplexity: 248320.0000   (vocab_size = 248320)
```

The BF16 source does this, so it is not a quantization effect — an NVFP4 export built from it behaves the same way.

## What it is not

The load path looks healthy in the log, which is what makes this quiet:

```
WARN  weight_map.cpp:1400: output projection (out_proj) was not found (may be tied to embed_tokens)
INFO  safetensors_loader.cpp:995: Tied output projection to token embedding
INFO  weight_upload.cu:949: Output projection shares GPU data with token embedding (weight tying)
INFO  weight_map.cpp:1320: WeightMap (qwen35): assigned 426 tensors, skipped 312, layers=32, experts=0
```

So the tying fallback *fires* and reports success. Tied embeddings alone are also not the trigger — **Qwen3-0.6B** (`tie_word_embeddings: true`) generates correctly on the same build.

## Narrowed

| model | architecture | `tie_word_embeddings` | vision | generates |
|---|---|---|---|---|
| Qwen3-0.6B | `Qwen3ForCausalLM` | **true** | no | ✅ |
| Qwen3.6-27B-Text-NVFP4 | `Qwen3_5ForConditionalGeneration` | false (real `lm_head`) | no | ✅ |
| Qwen3.6-35B-A3B-NVFP4 | `Qwen3_5MoeForConditionalGeneration` | false | yes | ✅ |
| **Qwen3.5-4B** | `Qwen3_5ForConditionalGeneration` | **true** | yes | ❌ |

The failing case is the only one that combines a `Qwen3_5*` architecture with tied embeddings. Under that architecture the tensor is named `model.language_model.embed_tokens.weight` (note the `language_model.` infix) rather than `model.embed_tokens.weight`, so the same fallback that works for `Qwen3ForCausalLM` may be tying a pointer that is not the tensor the LM head path actually reads.

One possibly related line from the same load:

```
INFO  pre_dequant_phase4_tensor_registry.cu:348: Phase-4 plan/actual MISMATCH: TOK_EMBED plan-tier=2 actual-tier=5
```

## Why it matters beyond one checkpoint

`PPL == vocab_size` is worth knowing as a diagnostic on its own: it means uniform logits, i.e. the head is dead, and it is distinguishable at a glance from "quantization made this worse". It is also **silent** — no error, no refusal, and a plausible-looking load summary.

This is the third instance of the class where a model loads, reports success, and serves garbage (cf. the vision tower loading 247/316 and answering anyway).

kekzl:
Fixes #1282.

## The bug

imp's GDN scan computes `g = exp(A_log * softplus(alpha + dt_bias))`, matching GGUF semantics — llama.cpp's converter pre-applies `-exp()` to the raw HF `A_log`, so the kernel expects the post-transform value. The upload path applies that transform for HF checkpoints, but decided *"is this HF?"* from the **dtype**, treating `F32` as "GGUF, already transformed".

An HF checkpoint can store `A_log` as F32. **Qwen3.5-4B does.** Its raw `A_log` reached the scan as a *positive* decay rate, so the recurrent state grew instead of decaying:

| token | 0 | 1 | 2 | 3 | 4 | 5 | 6 |
|---|---|---|---|---|---|---|---|
| scan absmax | 0.04 | 0.06 | 0.40 | 2.51 | **110.06** | **31680.00** | **inf** |

It crosses FP16 between t5 and t6; `rmsnorm(inf)` yields NaN and every layer past 11 is non-finite. The model **loaded without an error** and emitted `<|im_start|>` forever, with perplexity 248320.0000 — exactly `vocab_size`, i.e. uniform logits.

## How it was isolated

Comparing the two load paths on the same model, layer 8:

| source | range | all negative |
|---|---|---|
| GGUF `blk.8.ssm_a` | -7.7437 … -0.0681 | yes |
| HF `linear_attn.A_log` | -2.6875 … **+2.0469** | no |
| computed `-exp(HF A_log)` | **-7.7437 … -0.0681** | matches GGUF |

Every *other* scan input was already identical across both paths — `gdn_conv_f32` agreed to four decimals — which is what pinned it to this one tensor.

## The fix

Decide from the **values**, which is decidable: `-exp(x)` is strictly negative for every real x, so any value `>= 0` can only be a raw HF `A_log`. The dtype test stays as an OR, since BF16/F16 storage is HF-only.

When every value is negative **and** the source is F32, this is a no-op — exactly the previous behaviour — so no existing GGUF model changes.

## Verification

| | before | after |
|---|---|---|
| Qwen3.5-4B BF16, PPL | **248320.0000** | **12.6697** |
| Qwen3.5-4B BF16, generation | `<\|im_start\|>` forever | coherent |
| Qwen3.5-4B mxfp4 **GGUF** (regression) | — | unchanged, still answers "… is **Paris**." |

`make dev-test` green (939 tests). `verify-fast` green: decode tg128 287.17 (-0.01% vs pin).

https://claude.ai/code/session_019Vsh8gvJbp8uVg2gvpkir8

### DIFF
--- a/src/exec/executor_workspace.cu
+++ b/src/exec/executor_workspace.cu
@@ -160,6 +160,35 @@ bool GraphExecutor::init(const Model& model, QType compute_dtype, bool use_pdl, 
              &logits_, &fp32_accum_buf_, &fp32_hidden_);
 
     // Compute shared workspace sizes (no allocation — deferred to allocate_workspaces()).
+    // Detect GDN layers (Gated DeltaNet, e.g., Qwen3.5) BEFORE the shared sizes
+    // are computed. GDN carries its input projection in FP32 for precision, so
+    // compute_shared_sizes() charges 4 bytes/elem for it — but only if it knows
+    // the model is GDN. Reading has_gdn_ while it is still false reserved HALF
+    // the bytes the pointer carve then handed out, and configure_ssm_workspace()
+    // ran later, when the flag WAS set, so the two disagreed by exactly 2x.
+    //
+    // On a GDN+MoE model the MoE phase dominates the shared arena and hid this;
+    // on a GDN+dense model (Qwen3.5-4B) the SSM phase is the maximum, and the
+    // overrun corrupted every token past the halfway point — NaN from the first
+    // recurrent layer whose block exceeded it, silently and prefill-only (#1282).
+    //
+    // Deliberately placed AFTER the max_tokens cap above, which reads has_gdn_
+    // while it is still false. That is the AS-BUILT behaviour the T2 reservation
+    // replicates on purpose (AUDIT B18) — moving this any earlier would change
+    // the cap and under-reserve the arena by 2x instead.
+    {
+        int gdn_idx = 0;
+        for (int i = 0; i < cfg.n_layers; i++) {
+            if (model_->layer(i).gdn_gate.data != nullptr) {
+                gdn_idx++;
+            }
+        }
+        if (gdn_idx > 0) {
+            has_gdn_ = true;
+            IMP_LOG_INFO("GDN layers: %d out of %d total", gdn_idx, cfg.n_layers);
+        }
+    }
+
     // Deferring GPU allocation maximizes VRAM available for expert weight upload.
     ws_.compute_shared_sizes(max_tokens_);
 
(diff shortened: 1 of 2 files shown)


kekzl:
Superseded: this change landed on main as part of the #1285 squash (3cff788e) — that PR was branched off this one, so its squash carried the A_log fix with it. `git diff main..fix/ssm-a-log-transform-f32` shows no difference in `src/model/weight_upload.cu`. Closing; no work is lost.

Stacking is what did this (repo rule says don't). Nothing to redo, but three logically separate fixes now share one commit message on main.

Issue #1251, opened 2026-08-05
Qwen3.6-35B-A3B UD-Q4_K_M: KV pool clamps to the 16-block floor (512 tokens), every generation cancelled at ~476 tokens


kekzl:
## What happened

`Qwen3.6-35B-A3B-UD-Q4_K_M.gguf` loads and serves, but its KV pool collapses to the **16-block
floor — 512 tokens** — under default config. Every generation is cancelled at ~476 completion
tokens with `finish_reason: "cancelled"` and an empty `content`. The model is effectively unusable
for anything longer than a short reply, even though nothing about the load fails.

No documented knob recovers it: lowering `runtime.max_seq_len`, pinning `--max-batch 1`, and
switching to `--kv-fp8` all produce the same 16 blocks. The requested pool is computed and then
clamped away — the log shows `KV cache: 128 -> 16 blocks`.

`docs/supported-models.md:49` lists this GGUF at 243 tok/s / 22 GB, which is not reachable at any
useful context under default config.

Found while investigating #1248 (using this GGUF as the counter-example for the NVFP4 checkpoint
of the same model). Unrelated to that issue's root cause — filing separately as discussed there.

## Reproducer

```bash
# 1) default config — 512-token KV pool
docker run --rm --gpus all -p 8099:8080 -v /home/kekz/models:/models \
  ghcr.io/kekzl/imp:latest \
  --model /models/qwen3.6-35B-A3B-gguf/Qwen3.6-35B-A3B-UD-Q4_K_M.gguf \
  --host 0.0.0.0 --port 8080

# 2) any generation past ~476 tokens is cancelled
curl -s http://localhost:8099/v1/chat/completions -H 'Content-Type: application/json' -d '{
  "model":"Qwen3.6-35B-A3B-UD-Q4_K_M.gguf",
  "messages":[{"role":"user","content":"Schreibe einen zusammenhaengenden Aufsatz von etwa 800 Woertern ueber die Geschichte der Dampfmaschine."}],
  "temperature":0.7,"stream":false,"max_tokens":1200}'
# -> finish_reason "cancelled", completion_tokens 476, content ""

# 3) none of these change the 16 blocks
#    --set runtime.max_seq_len=8192
#    --max-batch 1 --kv-fp8 --set runtime.max_seq_len=4096
```

## Environment

[...]

PR #1286, opened 2026-08-06
fix(runtime): say it out loud when the KV pool cannot hold the requested context

kekzl:
#1251 made the floor case loud: when nothing is left to size the pool from, the load warns instead of reporting success and then cancelling every generation at admission. The quiet half stayed quiet — a pool that is a real size, well above the floor, and still smaller than one `max_seq_len` sequence. Same outcome for the operator, no message.

The hard failure that already names this arithmetic only fires with an explicit `--vram-budget`. Without one the sizing stays best-effort by design, so this warns rather than fails.

## Only for an operator-set max_seq_len

An auto value is a projection this clamp is *expected* to undercut — `init_compute_max_seq_len_` sizes the GGUF path from raw free VRAM on purpose and leaves the overshoot "absorbed by the downstream KV clamp". Warning there would fire on healthy loads and bury the case that is a fault. Where the value came from is now recorded rather than inferred, since the value itself no longer says once auto has filled it in.

That distinction is the whole design of this PR; without it the warning is noise.

## Testable without a GPU

`kv_pool_verdict()` carries the decision, for the same reason `kv_blocks_from_residual` moved out of the engine in #1251: inline, nothing could reach it without a GPU and a 22 GB checkpoint. Six tests, including the boundary that decides whether a pool sized to exactly the request counts as a fault (it does not).

Mutation-validated — each mutation caught by exactly the intended test:

| Mutation | Caught by |
|---|---|
| `blocks < need` → `<=` | `ExactlyOneSequenceIsSufficient` |
| `floored` early-return removed | `FlooredKeepsItsOwnMessage` |
| ceiling → floor division | `PartialTrailingBlockCounts` |

`ctest -L unit` green (test-core: 949 tests, 889 passed, 60 GPU-gated skips).

## What this does not cover

The auto path can still advertise a `max_model_len` that admission will cancel — `/v1/models` publishes `max_seq_len`, and the pool is clamped after that value is fixed. That is a separate question (correct the projection vs. advertise the measured capacity) and deliberately not decided here.

https://claude.ai/code/session_019Vsh8gvJbp8uVg2gvpkir8

### DIFF
--- a/src/memory/vram_query.h
+++ b/src/memory/vram_query.h
@@ -113,6 +113,38 @@ struct KvResidualSizing {
 // plan is a projection, the residual is the truth) from "there was nothing
 // left to size from" (always an operator-visible fault), so the caller can
 // be quiet about the first and loud about the second.
+// Blocks one max_seq_len sequence occupies. 0 when either input is unset,
+// which reads as "no requirement to check against".
+inline int kv_blocks_per_sequence(int max_seq_len, int block_size) {
+    if (max_seq_len <= 0 || block_size <= 0)
+        return 0;
+    return (max_seq_len + block_size - 1) / block_size;
+}
+
+// What the operator has to be told about the pool this sizing produced.
+//
+// `Floored` has had its own message since #1251: nothing was left to size
+// from. `ShortOfOneSequence` is the quiet half of the same fault and had
+// none — the pool is a real size, just too small for a single max_seq_len
+// request, so the load reports success and every full-length generation is
+// cancelled at admission instead. Both are operator faults; only the first
+// was audible.
+enum class KvPoolVerdict {
+    Sufficient,          // holds at least one full-length sequence
+    ShortOfOneSequence,  // sized, but no full-length request can be admitted
+    Floored,             // nothing was left to size from
+};
+
+inline KvPoolVerdict kv_pool_verdict(const KvResidualSizing& sizing, int max_seq_len,
+                                     int block_size) {
+    if (sizing.floored)
+        return KvPoolVerdict::Floored;
+    const int need = kv_blocks_per_sequence(max_seq_len, block_size);
+    if (need > 0 && sizing.blocks < need)
+        return KvPoolVerdict::ShortOfOneSequence;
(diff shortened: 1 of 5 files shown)


kekzl:
## End-to-end on a real load — the engine-side branch no test can reach

Ornith-1.0-35B-NVFP4 on the 5090, `build-dev` binaries, five arms. The warning has to fire in exactly one of them; a warning that also fires in the other four would be the noise this PR is built to avoid.

| Arm | `max_seq_len` | KV pool | warns |
|---|---|---|---|
| A | explicit 32768 | 1024 blocks (32768 tokens) | no |
| B | explicit 262144 | 8192 blocks (262144 tokens) | no |
| C | **auto** | 4096 blocks (131072 tokens) | no |
| D | explicit 4096 | 128 blocks (4096 tokens) | no |
| **E** | **explicit 2097152** | **11677 blocks (373664 tokens)** | **yes** |

```
[WARN] engine_kv_cache_init.cpp:411: KV cache: the pool ends up at 11677 blocks
(3649 MiB, 373664 tokens) but the requested max_seq_len=2097152 needs 65536 blocks
(20480 MiB). Every full-length request will be cancelled at admission even though
this load reports success. Lower --max-seq-len, lower the weight-cache demand
(moe.reserve_mib, --kv-fp8), or free at least 16831 MiB for the KV pool.
```

The arithmetic checks out against the load itself: 20480 − 3649 = 16831 MiB, and the pool reported at line 453 (11677 blocks / 3649.06 MiB) is the one the message names.

Arm D is the boundary in the field — a pool sized to exactly the request, which the unit test `ExactlyOneSequenceIsSufficient` pins as *not* a fault. Arm C is the auto path this PR deliberately stays quiet on.

## Ordering, since a flag read before it is written would be a silent no-op

`init_compute_max_seq_len_()` runs at `engine.cpp:647`, `init_kv_cache()` at `engine.cpp:787`. `max_seq_len_explicit_` is set before it is read — checked rather than assumed, because a gate standing in front of the check it guards is the shape of the bug this PR closes.

Issue #654, opened 2026-06-10
flash_attention_blackwell (final fallback) produces garbage end-to-end — teacher-forced PPL 11.9M on gemma-3-12b


kekzl:
Found during the #511 fp8-FMHA investigation (2026-06-10).

Forcing the final-fallback kernel (`--prefill-chunk-size 0 --set attention.fmha_prefill_threshold=1 --set attention.fp8_fmha=never --set attention.fmha_sm120=never`) on gemma-3-12b Q4_K_M, 8274-token corpus:

| path | teacher-forced PPL |
|---|---|
| cuBLAS chunked reference | 16.56 |
| FP16 WMMA (`fmha_sm120_prefill`) | 11.12 |
| **`flash_attention_blackwell`** | **11,923,036** |

The kernel is unreachable in the default config (FA2/WMMA serve everything first), but it is still the chain's last resort: any config the FP16 WMMA kernel declines (hd % 16 != 0, smem-constrained `skip_batch_dequant` fallback) lands here and silently produces garbage. Either fix it, or remove it and make the chain decline loudly (route to cuBLAS / error) — a garbage last-resort is worse than no fallback. Candidates list in the 06-09 structure audit already flags it as a dead-kernel candidate (C8).

Repro is the command above; per-position NLL via `IMP_PPL_DUMP=1`.

kekzl:
Fixed by PR #658 (merged): flash_attention_blackwell returns a launch-checked decline, the silent tc fallback is removed, and attention_prefill_dispatch throws on chain exhaustion. Root cause was an unchecked launch failure (smem over limit at hd=256), not kernel math — forced-blackwell hd=128 reads PPL 40.50 vs cuBLAS 40.51.

kekzl:
Fixed by PR #658 (merged): flash_attention_blackwell returns a launch-checked decline, the silent tc fallback is removed, and attention_prefill_dispatch throws on chain exhaustion. Root cause was an unchecked launch failure (smem over limit at hd=256), not kernel math — forced-blackwell hd=128 reads PPL 40.50 vs cuBLAS 40.51.

PR #1479, opened 2026-08-19
attn/gdn: three sweep findings — two real, one refuted

kekzl:
Three findings from the `find-stubs` sweep, each verified at the code before
anything was changed. Two turned out to be real, one did not — recorded here so
it is not re-chased.

## 1. `DISABLED_BasicHD256` blamed the wrong side

The test carried *"HD=256 requires large shared memory; disabled pending smem
optimization"*, which reads as a defect in the kernel under test. Running it
says otherwise:

```
[INFO] FMHA MXFP4 sm120 ACTIVE (hd=256, Bq=32, blockscale=0, ...)
tests/test_attention_fmha_mxfp4.cu:123: Value of: flash_attention_blackwell(...)
  Actual: false
```

The MXFP4 launcher accepts hd=256 — its layout needs ~80 KiB of the 99 KiB
opt-in at Bq=32. What returns false is the test's **FP16 reference**:
`flash_attention_blackwell` cannot serve hd=256 on sm_120 at all (Br=64 needs
~176 KiB) and declines by design, as its own comment and
`ATTENTION_DISPATCH.md` tier 5 both already said.

Production has no such gap — the FA2 tier serves hd=256 and sits **ahead** of
the Blackwell tier in `attention_prefill_dispatch`. The reference now falls back
to `fmha_sm120_prefill` when Blackwell declines, and Blackwell still runs
wherever it ran before, so every previously passing case compares against
exactly what it compared against before.

| | max_err | mean_err | limits |
|---|---:|---:|---|
| hd=128 (already enabled) | 0.0096 | 0.000792 | 0.5 / 0.1 |
| **hd=256 (was disabled)** | **0.0145** | **0.001006** | 0.5 / 0.1 |

35/35 green. A live tier gained the coverage it was assumed not to have.

## 2. `gemm_grouped_dispatch` was dead

Zero callers in `src`, `tools` or `tests` — the code graph and a full-text
search agree — while its four `... tier not implemented` branches gave a
complete-looking API to a function nothing had ever called. Removed. Its live
sibling `gemv_dispatch` has three production call sites and keeps them;
`gemm_moe_batched`, which the dead function named, has seven of its own and is
untouched.

## 3. The GDN scan that returned an empty answer

`gdn_scan_fused_fp32out` had a fallback for head_dim/state_size combinations
without a kernel: it ran the FP16 decode kernel into a scratch buffer, logged
`fallback not implemented`, freed the scratch — and left `y_fp32` exactly as the
caller passed it in. The scan contributed nothing, the step completed, and the
only trace was one warning per token. Same shape as #654.

It throws now. Nothing reaches it today (every staged GDN checkpoint is
HD=SS=128 — Qwen3.5/3.6/3.8 `linear_key_head_dim` = `linear_value_head_dim` =
128), which is exactly why it could sit unfinished. Mamba2 checkpoints whose
shapes do differ (Nemotron-3.5: `mamba_head_dim` 64, `ssm_state_size` 128) use
the Mamba2 scan, not this one. The FP16 sibling `gdn_scan_fused_f32` keeps its
fallback — it writes straight into `y`, so that one is an implementation, not a
stub.

## Checked, not changed

The `CUTLASS_NVFP4` branch inside `gemv_dispatch` logs an error and returns
without writing `y`, but it is unreachable from the executor: that switch gives
`CUTLASS_NVFP4` its own case, so `default:` — the only path into `gemv_dispatch`
for an unhandled tier — never sees it.

## Verification

`make dev-test` green · full GPU suite green via the pre-commit gate on both
commits · MXFP4 attention 35/35 · WeightDispatch 11/11 · GDN/SSM 19 passed,
1 skipped (a microbench), unchanged · `docs_lint` green.

https://claude.ai/code/session_01Vg4uBk3ob7S9mPeTkNz8Pb


kekzl:
`make verify-fast` run explicitly on `ba06f1f1` (the backgrounded push printed no hook output, and exit 0 does not prove the gate ran — so it was re-run in the foreground rather than assumed):

```
decode  tg128 =  289.25 tok/s  (baseline  287.19, +0.72%)   PASS
prefill pp512 = 12587.82 tok/s (baseline 12406.87, +1.46%)  PASS
prefill pp4096= 15575.41 tok/s (baseline 15324.70, +1.64%)  PASS  (FA2)
own peak VRAM = 20718 MiB      (baseline 20716,   +0.01%)   PASS
graphs ON tg256 = 461.84 vs OFF 229.02 = 2.02x              PASS
smoke prompt (Qwen3-4B Q8_0)                                PASS
=== verify fast: OK ===
```

Median of 3 independent runs, spread 0.06%, clocks steady (sm 2902 MHz, mem 13801 MHz, 502 W peak).

PR #1497, opened 2026-08-19
weightmap: the MTP sidecar is not an unrecognised weight

kekzl:
## What

`WeightMap: unrecognised weight name` fired **270 times** on
Nemotron-3.5-Lightning for an MTP head that loads correctly. The head is a
sidecar with its own loader and its own naming (`mtp.layers.N.*`); it is not a
weight of the main model, so it is not a finding here.

## Why it matters more than log volume

That string is what you grep for after a first load, to catch a checkpoint whose
tensors this map cannot read. It is the check that found the 869 silently
skipped tensors in a community NVFP4 export. 270 false positives make it useless
exactly where it is needed.

The 270 were also counted as `skipped`, and that count is printed inside the
"not one expert tensor was recognised" error.

## What it uncovered

With the sidecar classified, the summary line reads:

```
WeightMap (nemotron_h_moe): assigned 18205 tensors, skipped 12,
                            270 MTP sidecar (own loader), layers=52, experts=128
```

and those **12 are a finding that was invisible before**:

```
model.layers.{5,12,19,26,33,42}.self_attn.k_proj.k_scale
model.layers.{5,12,19,26,33,42}.self_attn.v_proj.v_scale
```

`k_scale` / `v_scale` for exactly the 6 attention layers this hybrid has. They
are KV-cache quantisation scales the checkpoint ships and imp never reads
(`grep -rn "k_scale" src/` finds no consumer; the FP8 path computes its own).
Whether adopting them changes output quality is unmeasured and not touched here.
It is now visible, which it was not while it sat behind 270 false warnings.

## Verification

On the model, before and after: **270 warnings to 0**, head still loading its
270 tensors (`MTP head loaded: ... (270 tensors, Nemotron layout ...)`).

No unit test. The only observable here is a log level, and a test asserting that
`"mtp.x"` starts with `"mtp."` would assert nothing, which is the failure mode
the `find-stubs` skill calls the most dangerous kind of unfinished. The
behaviour is pinned by the model run above instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,10 +35,6 @@ there instead of retelling it.
   again.** A token carrying string content *and* the closing quote got neither
   category, so the pre-filter dropped it before the FSM. Costs ~5.6 % decode.
 
-- **A checkpoint's MTP tensors are no longer reported as unrecognised weights.**
-  They have their own loader; 270 false WARN lines on Nemotron-3.5 were hiding
-  12 real ones (`k_scale`/`v_scale`, which imp does not read).
-
 - **`response_format: regex` and `grammar` answer the request instead of
   `!!!!...` until `max_tokens`.** The allow list was uploaded at the width of the
   logits row into a tokenizer-sized buffer. Broken since #1091/#1095.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,10 @@ there instead of retelling it.
   again.** A token carrying string content *and* the closing quote got neither
   category, so the pre-filter dropped it before the FSM. Costs ~5.6 % decode.
 
+- **A checkpoint's MTP tensors are no longer reported as unrecognised weights.**
+  They have their own loader; 270 false WARN lines on Nemotron-3.5 were hiding
+  12 real ones (`k_scale`/`v_scale`, which imp does not read).
+
 - **`response_format: regex` and `grammar` answer the request instead of
   `!!!!...` until `max_tokens`.** The allow list was uploaded at the width of the
   logits row into a tokenizer-sized buffer. Broken since #1091/#1095.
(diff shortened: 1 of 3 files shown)


kekzl:
Follow-up on the 12 tensors this uncovered: they are not a Nemotron quirk.
Scanning the `model.safetensors.index.json` of every local checkpoint for
`*.k_scale` / `*.v_scale`:

```
 96  Qwen3-Coder-30B-A3B-Instruct-FP4
 96  Qwen3-30B-A3B-NVFP4-Modelopt
 96  Gemma-4-12B-NVFP4
 80  Qwen3-14B-NVFP4
 80  Phi-4-reasoning-plus-NVFP4
 12  NVIDIA-Nemotron-3.5-Lightning-30B-A3B-NVFP4
```

Six checkpoints ship calibrated KV-cache scales and imp reads none of them; the
FP8 KV path computes its own. Whether adopting them moves output quality is
unmeasured, and measuring it needs the card, which is busy with the release
suite runs. Recording it here so it is not lost.

Nemotron shows only 12 because the hybrid has 6 attention layers out of 52; the
dense models show one pair per layer.

kekzl:
Two additions from review.

**The risk this diff introduces**, which belongs in the body rather than a test:
a genuinely unknown tensor whose name happens to start with `mtp.` is now
swallowed at DEBUG instead of warned about. That is the inversion of the very
detector this repairs. It is narrow today (the MTP loader owns that prefix
entirely), but anyone widening this branch should widen the classification, not
the silence.

**And the finding no longer lives only in a log line.** `docs/LIMITATIONS.md`
now carries the KV-scale gap under "Known-bad and known-limited behaviour": the
tensor names, the count per model, and the fact that the FP8 path derives its
own scales and adopting the shipped ones is unmeasured. A log line is where this
would have gone invisible again, which is the same mechanic that made 270 false
warnings a problem to begin with.

PR #1516, opened 2026-08-21
bench: re-measure the llama.cpp lead against b10524, pinned by digest

kekzl:
The competitive sweep in `docs/BENCHMARKS.md` was 2026-07-12 against llama.cpp
build 9976, and it recorded the **tag**, not the digest. `:full-cuda` moves, so
nothing in the repo said which competitor build a published lead was measured
against. `make bench-competitive` now runs `scripts/bench_competitive.sh`, which
pins the image by digest.

## Result: release bar 2 holds on all four heroes

| Model (shared quant) | imp default | imp spec-off | llama.cpp b10524 | lead | lead 07-12 |
|---|---:|---:|---:|---:|---:|
| Qwen3-8B Q8_0 | 396.76 | 284.60 | 159.74 | **+148 %** | +48 % |
| Qwen3-14B Q6_K | 162.06¹ | 163.38 | 112.83 | **+44 %** | +42 % |
| Qwen3.6-35B-A3B UD-Q4_K_M | 283.99 | 284.48 | 220.53 | **+29 %** | +18 % |
| Gemma-4-26B-A4B UD-Q4_K_M | 244.16 | 244.24 | 210.34 | **+16 %** | +21 % |
| Qwen3-30B-A3B Q4_K_M (non-hero) | 314.62 | 314.64 | 303.28 | +3.7 % | +1.7 % |
| gpt-oss-20b MXFP4² | 412.41 | 412.09 | 330.09 | **+25 %** | +13-19 % |

**llama.cpp did not erode the lead.** b10524 is flat to slightly slower than
b9976 on every shared model (8B 160.5 to 159.74, 14B 115.3 to 112.83, 35B 226.5
to 220.53, Gemma-4 216.0 to 210.34, 30B 319.2 to 303.28).

## Two findings this reports rather than fixes

- **Gemma-4-26B-A4B is imp's one regression**, 261.3 to 244.16 (-6.6 %) against
  llama.cpp's -2.6 % on the same model. That is what took the lead from +21 % to
  +16 %.
- **The n-gram drafter never engages on Qwen3-14B Q6_K**: `verify_steps=0
  miss_steps=72 drafted=0` over a full bench run.

## Why there are two imp columns

imp's `--bench` prompt is self-repetitive, so where the drafter engages it
accepts essentially every draft (504 of 504 on Qwen3-8B) and `llama-bench` has
no equivalent. One number cannot mean both things. The pair also works as the
sweep's own repeatability control: where speculation is inert the two columns
must agree, and on 35B, Gemma-4, 30B and gpt-oss they agree to **0.2 %**.

¹ That control caught the one row where they did not. The sweep produced 154.77,
5.3 % below its own spec-off column; two isolated re-measurements gave 162.08 and
162.04 and the table carries their median. The first imp run after a 16 GiB
competitor model unloads is not on a settled card, so the settle between arms
was raised from 5 s to 20 s. This sweep's numbers were taken at 5 s.

² Basis changed: the 07-12 row compared imp on SafeTensors against llama.cpp on
GGUF. That checkpoint is no longer on this host, so both engines here read the
same GGUF. More apples-to-apples than the row it replaces, but not the same
measurement.

Docs, a script and a make target. No code paths touched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,17 @@ there instead of retelling it.
   member variables, because the 128 MiB pair above allocates and frees in
   different translation units and no per-file check can see it.
 
+### Changed
+
+- **The MTP truncation is documented as one outcome of ordinary speculative
+  divergence, not a defect with a location.** All six probe prompts diverge from
+  the non-speculative answer between byte 48 and byte 271, and the two that
+  diverge earliest after the truncating one produce full, clean answers, so the
+  divergence point carries no signal. The hidden-state diff that would localise it
+  is not obtainable: `diagnostics.dump_hidden_dir` is host-side and decode is
+  graph-replayed, giving 5 dump steps for both 40 and 200 generated tokens.
+  ([`LIMITATIONS.md`](docs/LIMITATIONS.md))
+
 ### Fixed
 
 - **A kernel whose only caller was its own test.** `fp32_accum_add_fp16_kernel` had
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,14 @@ there instead of retelling it.
 
 ### Added
 
+- **`make bench-competitive` re-runs the llama.cpp comparison with the competitor
+  image pinned by digest.** The previous sweep was six weeks and 548 upstream
+  builds old and compared against a build the repo did not record. Re-measured
+  against b10524: imp leads on all four heroes (+148 %, +44 %, +29 %, +16 %) and
+  llama.cpp is flat to slightly slower than b9976 on every shared model. It also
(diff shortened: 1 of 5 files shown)


kekzl:
Superseded by a rebased branch. This PR was born conflicting with `main`: it was branched off `2a049185` while #1515 was still in flight, and #1515 merged 29 minutes before this opened, touching the same `[Unreleased]` block. GitHub computes `pull_request` workflows against a merge ref it cannot build when there are conflicts, so no CI run and no Auto-merge run ever fired and the PR read as "checks pending" indefinitely. Force-push is gated on this repo, hence a new branch rather than an update in place.

PR #1519, opened 2026-08-21
gate(memory): invariant I2 was stated, counted, and unobservable

kekzl:
"Nothing allocates device memory while the engine is serving" had a counter that
could not see the traffic. `steady_state_allocations()` only observes what routes
through `Backend`, and the `--wrap` interposer that would see the rest compiles
under `IMP_ALLOC_INTERPOSE`, which appeared in **no make target and no CI job**.
So the number read zero in every shipping build no matter what the 469 direct
allocation sites outside `src/memory/` did. Ledger item 3.

`make check-alloc-interpose` builds it, starts a server at `max_batch_size=4`
with NVFP4 residual KV and an MTP chain, drives 20 requests four at a time, stops
the container with a 60 s timeout so the static-destructor report actually runs,
and reads the log.

## First run: 46 device allocations while serving

**27 were not a violation.** All of them one function, `mtp_workspace_allocate`.
`set_alloc_phase(Serving)` fires at the end of engine init while
`imp_enable_mtp_spec_decode` is a context-setup step the API exposes *after*
`imp_context_create`, and the server calls it before a single request. Init
traffic on the wrong side of the boundary. Labelled with `AllocPhaseScope`, not
exempted: the scope logs its reason at INFO, so an embedder that calls this
mid-serving is visible rather than masked.

**The remaining 19 are real, and they are five named sites**, not an aggregate:

| calls | bytes | site |
|---|---|---|
| 15 | ~0 | `try_launch_async_graph_loop`, `engine_graph_decode.cpp:408-412`, `:434`, allocated and torn down **per request** |
| 2 | 128 MiB | `run_attention` -> `chunk_eager_k_`/`_v_`, `executor_attention_prefill.cu:176-177` |
| 1 | ~0 | `banned_tokens_device_`, `engine_graph_decode.cpp:29`, lazy upload of a list known at init |
| 1 | 0.001 MiB | `VRAMAllocator::allocate`, the arena growing after the phase flip |

`PINNED_CALLS=19` fails in **both** directions: a rise is a new serving
allocation, a fall means someone fixed one and left the pin stale. Negative
control run at 18 and at 20, both red. Each site is named in the script and in
[`DEBT_LEDGER`](docs/audit/DEBT_LEDGER_2026_08_21.md) section (g), so the pin is
a work queue with addresses rather than a blessing. They are not fixed here:
four of the five are reworks in the decode and prefill launch paths, and each
wants its own measured PR.

## Two defects in the reporting path, and the second was mine

**The violation report was `IMP_LOG_DEBUG`.** Someone who opted into the
measurement build still saw nothing at default log level. It is `WARN` now, the
clean line is `INFO`, and the gate asserts one of the two appears **at all**:
absent both, the binary was built without the flag and a grep for violations
passes for the wrong reason.

**The gate first printed PASS with 19 violations in the log it had just read.**
The report's banner and its first class shared a line (no `\n` after
`while serving:`), and the parser anchored at the start of a line, so it skipped
`cudaMalloc` entirely and summed only the async and pinned rows. Both halves are
fixed: the format has its newline, and the parser matches the class name anywhere
in the line rather than depending on that.

The gate also refuses to judge before it can prove it reached the path it claims
to exercise: absent `residual buffer enabled` in the log, it fails.

## Also

The `shipping-prs` skill gains the #1516 triage lesson: `gh pr checks` reporting
nothing is a symptom, not "pending", and `mergeStateStatus` is the field that
answers it. A conflicted PR gets no workflow run at all and looks identical to a
queue backlog.

### DIFF
--- a/.claude/skills/shipping-prs/SKILL.md
+++ b/.claude/skills/shipping-prs/SKILL.md
@@ -53,7 +53,7 @@ git log -1 --stat origin/main                  # confirm your final commit is in
 
 **Don't branch a new topic off `main` while a previous PR's auto-merge is still in flight** — it squashes onto `main` any moment and your new branch misses it (conflict/rework later). Wait for the merge, `git pull --ff-only`, then branch.
 
-### `mergeStateStatus=BLOCKED` — triage before assuming
+### A PR that will not merge: triage before assuming
 
 Don't guess. Dump the real state first:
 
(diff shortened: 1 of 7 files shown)


kekzl:
Superseded by a rebased branch, same conflict shape as #1516: branched off `main` before #1518 merged, and #1518 touched the same `.PHONY` line, so this was born `DIRTY` and got no workflow run. Reopened from `gate/i2-interpose-v2`; force-push is gated here, hence a new branch rather than an update in place.

PR #1695, opened 2026-08-22
fix(server): what one request may cost, and two limits that keyed on what the client wrote

### LINKED ISSUE #1614: [SEC][S1] the rate limit keys on a client-supplied X-Forwarded-For and never evicts
## Fact
- Per-IP rate limiting keys on the client-supplied `X-Forwarded-For` header whenever it is present, with no trusted-proxy configuration, so a client bypasses the limit by varying one header.
- `tools/imp-server/main.cpp:180`

## Evidence
```
tools/imp-server/main.cpp:178
        // Rate limiting (per-IP, inference endpoints only)
        if (state.rate_limit > 0 && is_inference_endpoint(req.path)) {
            std::string ip = req.get_header_value("X-Forwarded-For");
            if (ip.empty())
                ip = req.remote_addr;
            if (!state.check_rate_limit(ip)) {
                res.status = 429;
```
- The header is preferred unconditionally: `req.remote_addr` is used only when the header is absent. There is no allowlist of proxy addresses and no `--trusted-proxy` flag:
```
$ grep -rn "trusted_proxy\|trust_proxy\|X-Real-IP\|remote_addr" tools/imp-server/
tools/imp-server/main.cpp:182:                ip = req.remote_addr;
```
- The same header is also used unvalidated as the identity written into the request log:
```
tools/imp-server/handlers_chat_params.cpp:48
    ctx.log_client_ip = req.get_header_value("X-Forwarded-For");
```
```
tools/imp-server/handlers_messages.cpp:307
    std::string log_client_ip = req.get_header_value("X-Forwarded-For");
```
```
tools/imp-server/handlers_responses.cpp:309
    std::string log_client_ip = req.get_header_value("X-Forwarded-For");
```
- The flag it defeats is documented as a limit:
```
tools/imp-server/args.cpp:52
            "  --rate-limit <n>      Max requests/minute per IP (default: 0=unlimited)\n"
```
- The tracker never evicts a key:
```
tools/imp-server/handlers.h:205
    bool check_rate_limit(const std::string& ip) {
        if (rate_limit <= 0)
            return true;
        std::lock_guard<std::mutex> lock(rate_mutex);
        auto now = std::chrono::steady_clock::now();
        auto cutoff = now - std::chrono::seconds(60);
        auto& stamps = rate_tracker[ip];
        // Remove old entries
        stamps.erase(std::remove_if(stamps.begin(), stamps.end(), [&](auto& t) { return t < cutoff; }),
                     stamps.end());
        if (static_cast<int>(stamps.size()) >= rate_limit)
            return false;
        stamps.push_back(now);
        return true;
    }
```
```
$ grep -n "rate_tracker" tools/imp-server/handlers.h tools/imp-server/*.cpp
tools/imp-server/handlers.h:211:        auto& stamps = rate_tracker[ip];
```
(one use site: insert-or-get. No `.erase(`, no size cap.)

## Impact
- `--rate-limit` is a no-op against any attacker: one random `X-Forwarded-For` value per request means every request is a fresh bucket. The guard holds only against a client that does not know the header exists.
- The same spoof poisons `--log-requests` attribution, so the JSONL audit trail names whatever IP the client chose.
- The spoof is also an unbounded-memory vector: `rate_tracker[ip]` inserts a new map entry per distinct header value and **no code path ever erases a map entry** — only the timestamps inside an entry are pruned. Each distinct value costs one `std::string` key (up to httplib's 8 KB header limit) plus a `std::vector`, and the map grows for the lifetime of the process.
- Blast radius: the seven paths in `is_inference_endpoint()` (`tools/imp-server/main.cpp:27-31`), i.e. all inference. `--max-concurrent` is unaffected (it reads queue depth, not the header).

## Repro
```bash
cd /home/kekz/github.com/kekzl/imp
sed -n '178,189p' tools/imp-server/main.cpp
grep -rn "trusted_proxy\|trust_proxy\|X-Real-IP" tools/imp-server/   # -> no output

# NOT RUN (runtime repro, needs a running server started with --rate-limit 5):
#   for i in $(seq 1 50); do curl -s -o /dev/null -w '%{http_code} ' \
#     -H "X-Forwarded-For: 10.0.0.$i" -X POST localhost:8080/v1/chat/completions \
#     -H 'Content-Type: application/json' \
#     -d '{"model":"m","messages":[{"role":"user","content":"hi"}],"max_tokens":1}'; done
#   expect: no 429 despite rate-limit 5. Repeat without the header: 429 from the 6th.
```

## Fix direction
- Default to `req.remote_addr`; honour `X-Forwarded-For` only when the peer address is in an operator-configured trusted-proxy list (`--trusted-proxy <cidr>`, repeatable).
- When trusted, take the **rightmost** untrusted entry of the header, not the whole string — the current code uses the raw header value, which for a chained proxy is a comma-separated list and keys on the whole list.
- Apply the same resolution to the three logging sites so the log and the limiter agree on one identity.

## Acceptance criteria
- [ ] With no `--trusted-proxy` configured, varying `X-Forwarded-For` does not change the rate-limit bucket; the 6th request at `--rate-limit 5` is a 429.
- [ ] With `--trusted-proxy` set and the request arriving from that address, the header is honoured.
- [ ] A test in `tests/api/` covers the spoof attempt, so `Real API contract` (`.github/workflows/ci.yml:341`) gates it.
- [ ] `--log-requests` records the resolved address, not the raw header.
- [ ] `rate_tracker` evicts entries whose stamp vector is empty, so its size is bounded by the number of active peers.

## Unverified
- Nothing. Both the header preference and the missing eviction are in the quoted lines.

### LINKED ISSUE #1615: [SEC][S1] /tokenize, /count_tokens and /admin escape the rate and concurrency limits
## Fact
- `is_inference_endpoint()` gates rate limiting and `--max-concurrent` by exact path match on seven routes; `/tokenize`, `/detokenize`, `/v1/messages/count_tokens`, `/admin/suspend` and `/admin/resume` are not among them and do real work per request.
- `tools/imp-server/main.cpp:27`
- `tools/imp-server/main.cpp:179`
- `tools/imp-server/main.cpp:192`

## Evidence
```
tools/imp-server/main.cpp:22
// The five inference routes that consume engine capacity and must go through
// rate-limiting / max-concurrent admission control. /v1/messages (Anthropic)
// and /v1/embeddings were previously omitted from these checks, silently
// bypassing both guards (the non-stream /v1/messages path reaches inference by
// directly calling handle_chat_completions() without re-entering pre-routing).
static bool is_inference_endpoint(const std::string& path) {
    return path == "/v1/chat/completions" || path == "/v1/completions" || path == "/v1/responses" ||
           path == "/v1/messages" || path == "/v1/embeddings" || path == "/v1/rerank" ||
           path == "/rerank";
}
```
- Both guards are conditioned on it:
```
tools/imp-server/main.cpp:179
        if (state.rate_limit > 0 && is_inference_endpoint(req.path)) {
```
```
tools/imp-server/main.cpp:192
        if (state.max_concurrent > 0 && is_inference_endpoint(req.path)) {
```
- The comparison is `==`, so the sub-path is a different string:
```
tools/imp-server/main.cpp:284
    svr.Post("/v1/messages/count_tokens", [&state](const httplib::Request& req, httplib::Response& res) {
        handle_count_tokens(req, res, state);
    });
```
  `"/v1/messages/count_tokens" == "/v1/messages"` is false, so `count_tokens` is exempt from both guards — while its own comment says it "tokenizes exactly like a real request" (`:282-283`).
- The unmetered routes, from `main.cpp`:
```
tools/imp-server/main.cpp:299  svr.Post("/tokenize", …)
tools/imp-server/main.cpp:303  svr.Post("/detokenize", …)
tools/imp-server/main.cpp:307  svr.Post("/admin/suspend", …)
tools/imp-server/main.cpp:311  svr.Post("/admin/resume", …)
```
- The per-request cost is real. `/v1/messages/count_tokens` builds the full prompt, applies the Jinja chat template and tokenizes; `/tokenize` runs BPE over the whole body. Both accept up to the global body cap:
```
tools/imp-server/main.cpp:145
    // Limit request body size to 100 MiB (prevents DoS via large base64 images)
    svr.set_payload_max_length(static_cast<size_t>(100) * 1024 * 1024);
```
- `--max-concurrent` is also a queue-depth read, not a semaphore, so it admits a burst before any request reaches the queue:
```
tools/imp-server/main.cpp:192
        if (state.max_concurrent > 0 && is_inference_endpoint(req.path)) {
            int queue = 0;
            {
                std::lock_guard<std::timed_mutex> lock(state.mtx);
                if (state.batching)
                    queue = state.batching->queue_depth();
            }
            if (queue >= state.max_concurrent) {
```

## Impact
- 100 MiB of text to `/tokenize` or `/v1/messages/count_tokens`, repeated without limit, is unbounded CPU and unbounded memory (the token vector plus the Jinja-rendered prompt) on a server whose operator set `--rate-limit` and `--max-concurrent` believing they were covered.
- `/admin/suspend` is unmetered too: a loop of suspend/resume thrashes the weights between host RAM and VRAM. Auth applies (both are behind `--api-key` when set), but not when it is unset — the default.
- The comment at `:22-26` records that this exact bug already shipped twice for `/v1/messages` and `/v1/embeddings`. The list is a hand-maintained enumeration with no test asserting that every registered POST route is classified.
- Blast radius: five routes, unauthenticated by default.

## Repro
```bash
cd /home/kekz/github.com/kekzl/imp
sed -n '27,31p' tools/imp-server/main.cpp
grep -n 'svr\.Post' tools/imp-server/main.cpp
# 12 POST routes registered; 7 appear in is_inference_endpoint().

# NOT RUN (runtime repro, server started with --rate-limit 2):
#   for i in $(seq 1 20); do curl -s -o /dev/null -w '%{http_code} ' -X POST \
#     localhost:8080/tokenize -H 'Content-Type: application/json' -d '{"text":"hi"}'; done
#   expect: twenty 200s, no 429.
```

## Fix direction
- Invert the list: gate every POST route by default and enumerate the *exempt* ones (`/health`, CORS preflight), so a newly added route is metered unless someone deliberately exempts it.
- Failing that, use a prefix match for `/v1/messages` and add the four missing paths.
- Add a test that walks the registered routes and asserts each is either classified as inference or explicitly listed as exempt — the enumeration has now drifted three times.
- Make `--max-concurrent` a semaphore held for the request's lifetime instead of a queue-depth sample.

## Acceptance criteria
- [ ] With `--rate-limit 2`, the third request to `/tokenize`, `/detokenize` and `/v1/messages/count_tokens` returns 429.
- [ ] A test enumerates every registered route and fails when a new one is neither metered nor explicitly exempt.
- [ ] `--max-concurrent 1` admits one concurrent inference request, not a burst.

## Unverified
- The exact per-request cost of `/v1/messages/count_tokens` at 100 MiB; the absence of any limit on it is proven.

### LINKED ISSUE #1616: [SEC][S1] n, documents and input multiply one request's work unbounded
## Fact
- Three request fields multiply the work of one HTTP request without any upper bound: `n` on chat completions, `documents` on rerank, `input` on embeddings. Each is one unit against `--max-concurrent` and `--rate-limit` regardless of the multiplier.
- `tools/imp-server/handlers_chat_params.cpp:107` (`n`, lower bound only)
- `tools/imp-server/handlers_chat_core.cpp:725` (`n` as a loop bound)
- `tools/imp-server/handlers_rerank.cpp:102`
- `tools/imp-server/handlers_misc.cpp:263`

## Evidence
- `n` is clamped from below and never from above:
```
tools/imp-server/handlers_chat_params.cpp:107
    ctx.params.n_completions = body.value("n", 1);
    if (ctx.params.n_completions < 1)
        ctx.params.n_completions = 1;
```
```
tools/imp-server/handlers_chat_core.cpp:724
    for (int ci = 0; ci < ctx.params.n_completions; ci++) {
        // For subsequent completions, create a new request and submit it
        if (ci > 0) {
            imp_req = build_imp_request_(ctx, saved_tokens, ci, /*stream=*/false);
```
  Each iteration is a full independent generation, run sequentially, each bounded only by the context window.
- The neighbouring field `max_tokens` **is** bounded, which is what makes the omission visible:
```
tools/imp-server/handlers_chat_core.cpp:638
    // Clamp max_tokens to remaining context window
    int remaining = ctx.snap.max_seq_len - ctx.snap.n_prompt_tokens;
    if (ctx.params.max_tokens > remaining)
        ctx.params.max_tokens = remaining;
```
```
tools/imp-server/handlers_rerank.cpp:91
    // Cohere and Jina both spell it "documents"; vLLM also accepts "texts".
    ...
    std::vector<std::string> documents;
    ...
            documents.push_back(d.get<std::string>());
        ...
            documents.push_back(d["text"].get<std::string>());  // Cohere's object form
```
```
tools/imp-server/handlers_misc.cpp:261
    // Collect inputs: "input" can be a string or array of strings
    ...
    if (body.contains("input")) {
        if (body["input"].is_string()) {
            inputs.push_back(body["input"].get<std::string>());
        } else if (body["input"].is_array()) {
            for (const auto& item : body["input"]) {
```
- No cardinality check on either:
```
$ grep -n "documents.size() >\|max_documents\|kMaxDocuments" tools/imp-server/handlers_rerank.cpp
(no output)
$ grep -n "inputs.size() >\|max_inputs\|kMaxInputs" tools/imp-server/handlers_misc.cpp
(no output)
```
- The only limits that apply are per-request, not per-element: the 100 MiB body cap (`tools/imp-server/main.cpp:146`) and `--max-concurrent`, which counts *requests*, not the work inside one:
```
tools/imp-server/main.cpp:192
        if (state.max_concurrent > 0 && is_inference_endpoint(req.path)) {
            int queue = 0;
            {
                std::lock_guard<std::timed_mutex> lock(state.mtx);
                if (state.batching)
                    queue = state.batching->queue_depth();
            }
            if (queue >= state.max_concurrent) {
```
- `top_n` is bounded to the array size, which is itself the unbounded quantity:
```
tools/imp-server/handlers_rerank.cpp:117
    int top_n = body.value("top_n", static_cast<int>(documents.size()));
    if (top_n <= 0 || top_n > static_cast<int>(documents.size()))
```

## Impact
- `{"n": 1000000}` on a non-streaming `/v1/chat/completions` runs a million full generations sequentially inside one request, each up to the context window. It is one unit against `--max-concurrent` and one against `--rate-limit`.
- A single `/v1/rerank` request carrying ~10^6 short documents inside the 100 MiB body counts as **one** against `--max-concurrent` and **one** against `--rate-limit`, while occupying the engine for the duration of a million cross-encoder passes. Both operator-facing limits are per-request and therefore describe nothing about the load.
- Same for `/v1/embeddings`. Closed issue #890 already recorded that `--max-input-tokens` is not enforced on `/v1/embeddings`; the array cardinality is the second half of that and is still open.
- No cancellation path is visible for a partially processed batch, so the request cannot be shed under pressure.
- Blast radius: `/v1/chat/completions` (non-streaming; the streaming path already refuses `n > 1` at `handlers_chat_params.cpp:112`), `/v1/rerank`, `/rerank`, `/v1/embeddings`. Unauthenticated by default.

## Repro
```bash
cd /home/kekz/github.com/kekzl/imp
sed -n '91,120p' tools/imp-server/handlers_rerank.cpp
sed -n '261,270p' tools/imp-server/handlers_misc.cpp
grep -n "max_documents\|kMaxDocuments\|max_inputs" tools/imp-server/handlers_rerank.cpp tools/imp-server/handlers_misc.cpp   # -> no output

# NOT RUN (runtime repro, needs a running server with a rerank model):
#   python3 -c 'import json;print(json.dumps({"model":"m","query":"q",
#     "documents":["d"]*1000000}))' > b.json
#   time curl -sS -X POST localhost:8080/v1/rerank -H 'Content-Type: application/json' --data @b.json
```

## Fix direction
- Cap `n` (OpenAI's own limit is 128) and return 400 past it.
- Cap the element count on both array endpoints (`--max-batch-items`, default a few hundred) and return 400 past it — consistent with `tools/imp-server/CLAUDE.md`: "a request the server cannot honour is a 4xx".
- Enforce `--max-input-tokens` per element as well as in aggregate, closing the other half of #890.
- Charge `--max-concurrent` by work admitted rather than by request, or document plainly that it counts requests.

## Acceptance criteria
- [ ] `{"n": 1000000}` returns 400 rather than starting a generation.
- [ ] A `/v1/rerank` request above the element cap returns 400 with the standard envelope.
- [ ] The same for `/v1/embeddings`.
- [ ] `--max-input-tokens` rejects an oversized element in both.
- [ ] `tests/api/` covers the cap on `n` and on both array endpoints.

## Unverified
- Whether the rerank loop is sequential or batched, i.e. the exact wall-clock cost per element. The absence of any cardinality limit is proven.

### LINKED ISSUE #1617: [SEC][S1] logit_bias is uncapped and costs a blocking memcpy per entry per token
## Fact
- `logit_bias` accepts an unbounded map from the request and the sampler walks it with one **blocking** `cudaMemcpy` device-to-host plus one async copy back per entry, on every decode step.
- `src/exec/executor.cu:149`
- `src/exec/executor.cu:155`

## Evidence
```
src/exec/executor.cu:148
    // Apply logit bias
    if (state.n_logit_bias > 0 && state.logit_bias != nullptr) {
        for (int i = 0; i < state.n_logit_bias; i++) {
            int32_t tid = state.logit_bias[i].first;
            float bias = state.logit_bias[i].second;
            if (tid >= 0 && tid < vocab_size) {
                float logit;
                IMP_CUDA_CHECK_LOG(
                    cudaMemcpy(&logit, logits_ptr + tid, sizeof(float), cudaMemcpyDeviceToHost));
                logit += bias;
                IMP_CUDA_CHECK_LOG(
                    cudaMemcpyAsync(logits_ptr + tid, &logit, sizeof(float), cudaMemcpyHostToDevice, stream));
            }
        }
    }
```
- The token id is bounds-checked (`tid >= 0 && tid < vocab_size`) — that part is correct. What is missing is any bound on `n_logit_bias`:
```
$ grep -rn "n_logit_bias" src/ tools/ | grep -viE "state.n_logit_bias > 0|for \(int i"
(the field is set from the request; no size cap appears at any site)
```
- The loop runs per decode step, so its cost multiplies by `max_tokens`.

## Impact
- A request with a `logit_bias` map of 100,000 entries performs 100,000 synchronous device-to-host round trips **per generated token**. Each `cudaMemcpy` D2H is a full stream synchronisation, so this serialises the GPU behind host latency and stalls every other request sharing the engine.
- `max_tokens` multiplies it: 100,000 entries x 1,000 tokens is 10^8 blocking round trips from one request.
- The 100 MiB body cap allows on the order of 10^6 map entries.
- Unauthenticated by default, and one such request is one unit against `--max-concurrent` and `--rate-limit`.
- Blast radius: every decode path that carries sampling state — `/v1/chat/completions`, `/v1/completions`, `/v1/responses`.
- Rated S1: measurable engine-wide slowdown from a single request, but no memory-safety consequence.

## Repro
```bash
cd /home/kekz/github.com/kekzl/imp
sed -n '148,162p' src/exec/executor.cu
grep -rn "logit_bias" tools/imp-server/handlers_chat_params.cpp | head

# NOT RUN (runtime repro, needs a running server and a model):
#   python3 -c 'import json;print(json.dumps({"model":"m",
#     "messages":[{"role":"user","content":"hi"}],"max_tokens":64,
#     "logit_bias":{str(i):1.0 for i in range(50000)}}))' > b.json
#   time curl -sS -X POST localhost:8080/v1/chat/completions \
#     -H 'Content-Type: application/json' --data @b.json
#   compare tok/s against the same request without logit_bias.
```

## Fix direction
- Cap the map size at admission (a few hundred entries) and return 400 past it.
- Replace the per-entry round trip with one upload of the (id, bias) pairs plus a single kernel that applies them — the pattern the rest of the sampler already uses.

## Acceptance criteria
- [ ] A `logit_bias` map above the cap returns 400.
- [ ] Under the cap, applying bias costs at most one H2D copy and one kernel launch per step: `grep -c cudaMemcpy` in that block is 0.
- [ ] Measured tok/s with a full-size `logit_bias` is within a few percent of the unbiased baseline.
- [ ] `tests/api/` covers the 400.

## Unverified
- The exact cost per entry, hence the tok/s delta. Settle with the timed curl above. The blocking copy per entry per step is proven by the quoted lines.

### LINKED ISSUE #1618: [SEC][S1] the 404 envelope dumps the raw request path and throws on bad UTF-8
## Fact
- The 404/error envelope embeds the client-supplied request path and method and serialises it with raw `.dump()`, although the file's own comment states the envelope "can't throw on bad bytes" because `dump_safe` is used — and `dump_safe` exists precisely to stop `json::type_error.316` on ill-formed UTF-8.
- `tools/imp-server/main.cpp:357`
- `tools/imp-server/main.cpp:362`
- `tools/imp-server/utils.cpp:6`

## Evidence
```
tools/imp-server/main.cpp:346
    svr.set_error_handler([](const httplib::Request& req, httplib::Response& res) {
        if (!res.body.empty())
            return httplib::Server::HandlerResponse::Unhandled;
        const bool not_found = res.status == 404;
        const std::string msg = not_found ? "Unknown endpoint: " + req.method + " " + req.path
                                          : "Request failed with status " + std::to_string(res.status);
        if (req.path.rfind("/v1/messages", 0) == 0) {
            json err = {{"type", "error"},
                        {"error",
                         {{"type", not_found ? "not_found_error" : "invalid_request_error"},
                          {"message", msg}}}};
            res.set_content(err.dump(), "application/json");
        } else {
            json err = {
                {"error",
                 {{"message", msg}, {"type", res.status >= 500 ? "server_error" : "invalid_request_error"}}}};
            res.set_content(err.dump(), "application/json");
        }
```
- The invariant the file claims, 25 lines above:
```
tools/imp-server/main.cpp:319
    // Global safety net: any exception that escapes a handler must become a
    // JSON error envelope, never a bare "500 Internal Server Error". Malformed
    // or invalid-UTF-8 client input surfaces as a json::exception (parse_error,
    // or type_error.316 when an error message echoes the offending bytes) — map
    // those to 400. Everything else is a genuine internal failure → 500, but
    // still with a JSON body. dump_safe (inside send_json_error) guarantees the
    // envelope itself can't throw on bad bytes.
```
  The two envelopes at `:357` and `:362` are built inline and never go through `send_json_error`, so that guarantee does not cover them.
- The helper that would have covered them:
```
tools/imp-server/utils.cpp:6
std::string dump_safe(const json& j) {
    // error_handler_t::replace: emit U+FFFD for ill-formed UTF-8 instead of
    // throwing json::type_error.316. Identical output for valid UTF-8.
    return j.dump(-1, ' ', false, json::error_handler_t::replace);
}
```
- Raw `.dump()` sites in `main.cpp`: `:186`, `:204`, `:227` (static strings, safe) and `:357`, `:362` (client-derived, not safe).

## Impact
- `GET /%FF` (or any percent-encoded byte that is not valid UTF-8) routes to 404, lands in the error handler, and `err.dump()` throws `json::type_error.316` while constructing the error response.
- The exception is raised *inside* the error handler, not inside a routing handler, so `set_exception_handler` at `:326` — which is what maps json exceptions to a clean 400 — is not in that call path.
- Unauthenticated: the error handler runs for unmatched routes, which never reach the pre-routing auth check's protected branch.
- Blast radius: every unmatched path on every deployment.

## Repro
```bash
cd /home/kekz/github.com/kekzl/imp
sed -n '319,326p;346,364p' tools/imp-server/main.cpp
grep -n "\.dump()" tools/imp-server/main.cpp
sed -n '6,10p' tools/imp-server/utils.cpp

# NOT RUN (runtime repro, needs a running server):
#   curl -sv 'http://localhost:8080/%FF'
#   curl -sv 'http://localhost:8080/v1/messages/%FF'
#   expect (if httplib percent-decodes before routing): no JSON body, connection
#   dropped or a bare 500, instead of the documented envelope.
```

## Fix direction
- Replace both `err.dump()` calls with `dump_safe(err)`, or route the error handler through `send_json_error` like every other error path.
- Add a lint or a test asserting no raw `.dump()` on a client-derived string in `tools/imp-server/`; the helper exists and is applied inconsistently.

## Acceptance criteria
- [ ] `grep -n "\.dump()" tools/imp-server/main.cpp` shows no site that embeds `req.path`, `req.method` or any other client-derived value.
- [ ] `GET /%FF` returns a well-formed JSON error envelope with status 404.
- [ ] The case is covered in `tests/api/` under the `Real API contract` job.

## Unverified
- Whether the vendored cpp-httplib percent-decodes the path before routing, which is what puts the raw byte into `req.path`. Settle with the two curls above, or with `grep -n "decode_url" _deps/httplib-src/httplib.h` in a configured build tree. If it does not decode, the finding shrinks to a latent inconsistency with the file's own stated invariant.

### LINKED ISSUE #1619: [SEC][S1] shipped compose and README expose an unauthenticated API on 0.0.0.0
## Fact
- The container entrypoint forces `--host 0.0.0.0` and has no `IMP_API_KEY` translation at all, so the shipped `docker-compose.yml` and the README quickstart both publish an unauthenticated inference API on every host interface with no env var that could turn auth on.
- `docker-entrypoint.sh:38`
- `docker-compose.yml:14`
- `README.md:219`

## Evidence
```
docker-entrypoint.sh:35
# Host — default 0.0.0.0 inside container
if [ -n "$IMP_HOST" ]; then
    args+=(--host "$IMP_HOST")
elif [ "$CMD" = "imp-server" ]; then
    args+=(--host "0.0.0.0")
fi
```
- The entrypoint translates 14 env vars and not the one that matters:
```
$ grep -c "IMP_" docker-entrypoint.sh
28
$ grep -n "API_KEY\|api-key" docker-entrypoint.sh
(no output)
```
- The shipped compose publishes on all interfaces and passes no `command`, so no flags reach the binary:
```
docker-compose.yml:14
    ports:
      - "${IMP_PORT:-8080}:${IMP_PORT:-8080}"
```
```
docker-compose.yml:24
    environment:
      - IMP_MODEL=${IMP_MODEL:-}
      - IMP_HOST=0.0.0.0
```
- The README quickstart is the same shape:
```
README.md:218
docker compose build imp-server
docker run --gpus all -v ./models:/models -p 8080:8080 \
  imp:latest --model /models/your-model.gguf
```
- The binary's own default is safe — the exposure is created entirely by the container layer:
```
tools/imp-server/args.h:16
    std::string host = "127.0.0.1";
    int port = 8080;
```
- The documented compose is a *different* file from the shipped one, and does add a key:
```
docs/DEPLOYMENT.md:20
    command: ["--model", "/models/your-model.gguf", "--api-key", "${IMP_API_KEY}"]
```
- Same compose file ships Grafana with the default credential and anonymous read enabled, also published on all interfaces:
```
docker-compose.yml:110
    ports:
      - "${GRAFANA_PORT:-3002}:3000"
    ...
      - GF_SECURITY_ADMIN_USER=${GRAFANA_USER:-admin}
      - GF_SECURITY_ADMIN_PASSWORD=${GRAFANA_PASSWORD:-admin}
      - GF_AUTH_ANONYMOUS_ENABLED=true
```

## Impact
- `docker compose up` yields: unauthenticated `/v1/chat/completions`, `/v1/embeddings`, `/tokenize`, `/admin/suspend` and `/admin/resume` on `0.0.0.0:8080`; `/metrics` unauthenticated by default even when a key is set (`tools/imp-server/main.cpp:174`); Grafana on `0.0.0.0:3002` with `admin`/`admin`; Prometheus on `0.0.0.0:9090` with `--web.enable-lifecycle`.
- `POST /admin/suspend` from anyone on the LAN parks the weights and makes inference answer 503 (`tools/imp-server/main.cpp:307`).
- An operator who wants auth in the shipped compose cannot get it by setting an env var — they must know to override `command:`, which the shipped file does not hint at.
- Blast radius: every containerised deployment, which is the project's primary distribution (`ghcr.io/kekzl/imp:latest`).
- Rated S1: no memory-safety consequence, but the default deployment contradicts the documented one and hands an unauthenticated model server to the local network.

## Repro
```bash
cd /home/kekz/github.com/kekzl/imp
grep -n "API_KEY\|api-key" docker-entrypoint.sh          # -> no output
sed -n '35,40p' docker-entrypoint.sh
sed -n '14,15p;24,27p' docker-compose.yml
sed -n '116,121p' docker-compose.yml
sed -n '218,221p' README.md

# NOT RUN (runtime repro): `docker compose up -d` then, from another host on
# the LAN, `curl -s http://<host>:8080/v1/models` and `curl -X POST
# http://<host>:8080/admin/suspend`.
```

## Fix direction
- Add `IMP_API_KEY` -> `--api-key` to `docker-entrypoint.sh` alongside the other 14 translations, and refuse to start with `--host 0.0.0.0` and no key unless `IMP_ALLOW_ANONYMOUS=1` is set explicitly.
- Bind the published ports to loopback in the shipped compose (`127.0.0.1:8080:8080`, same for Grafana and Prometheus), matching what `docs/DEPLOYMENT.md` tells operators to do.
- Give Grafana a required password (no `:-admin` default) and drop `GF_AUTH_ANONYMOUS_ENABLED`.

## Acceptance criteria
- [ ] `IMP_API_KEY=x docker run … imp:latest --model …` starts with auth enabled; `curl` without the key gets 401.
- [ ] `docker compose up` with no env file does not accept an unauthenticated request from another host.
- [ ] `docker compose up` does not start Grafana with a default password.
- [ ] The shipped `docker-compose.yml` and the snippet in `docs/DEPLOYMENT.md:16-31` agree on the auth posture.

## Unverified
- Nothing about the configuration. Whether a given user's Docker install publishes to the LAN depends on their firewall; the compose file itself requests all interfaces.

### LINKED ISSUE #1622: [SEC][S2] HTTP connection limits are inherited httplib defaults, not configuration
## Fact
- imp configures no read timeout, no write timeout, no idle interval, no keep-alive cap and no thread-pool size on its HTTP server; every connection-level limit is whatever the build-time-fetched cpp-httplib happens to default to, and no flag exposes any of them.
- `tools/imp-server/main.cpp:99`
- `tools/imp-server/main.cpp:146`
- `cmake/imp-deps.cmake:12`

## Evidence
```
$ grep -rn "set_read_timeout\|set_write_timeout\|set_idle_interval\|set_keep_alive_max_count\|new_task_queue\|set_socket_options" tools/ src/
(exit 1, no output)
```
- The complete server-level configuration:
```
tools/imp-server/main.cpp:99
    httplib::Server svr;
    if (svr.bind_to_port(args.host, args.port) == 0) {
```
```
tools/imp-server/main.cpp:145
    // Limit request body size to 100 MiB (prevents DoS via large base64 images)
    svr.set_payload_max_length(static_cast<size_t>(100) * 1024 * 1024);
```
- The library is not vendored — it is fetched at build time from a pinned tag, so the effective limits are not readable from this repository at all:
```
cmake/imp-deps.cmake:12
set(IMP_DEP_HTTPLIB_TAG       v0.53.0  CACHE STRING "cpp-httplib git tag")
```
```
CMakeLists.txt:504
    FetchContent_Declare(httplib
        GIT_REPOSITORY https://github.com/yhirose/cpp-httplib.git
```
```
$ find . -name httplib.h -not -path './build*'
(no output)
```
- `--request-timeout` is a generation budget inside the handler, not a socket timeout, and is the only timeout the operator can set:
```
tools/imp-server/args.h:36
    int request_timeout = 300;      // --request-timeout: per-request timeout in seconds (0=unlimited)
```
- For scale, an unrelated cpp-httplib copy on this host (**v0.40.0, NOT imp's pinned v0.53.0 — informative only**) defaults to `CPPHTTPLIB_SERVER_READ_TIMEOUT_SECOND 5`, `CPPHTTPLIB_SERVER_WRITE_TIMEOUT_SECOND 5`, `CPPHTTPLIB_KEEPALIVE_MAX_COUNT 100`, `CPPHTTPLIB_THREAD_POOL_COUNT` derived from `hardware_concurrency`. So the server is probably not defenceless — but nothing in imp says so, nothing pins it, and nothing notices if a dependency bump changes it.

## Impact
- A dependency bump can silently change imp's exposure to connection-level attacks, and no test or gate would see it. The tag in `cmake/imp-deps.cmake:12` is the only pin, and it pins the code, not the behaviour.
- A per-read timeout does not stop a trickle: a client that sends one header byte inside every timeout window holds a worker thread indefinitely while `CPPHTTPLIB_HEADER_MAX_COUNT` allows 100 headers. The pre-routing guards (rate limit, `--max-concurrent`, auth) never run, because they fire after the request is fully received.
- The operator has no lever: there is no `--read-timeout`, no `--threads`, no `--keep-alive-max`. `docs/DEPLOYMENT.md:65-90` covers TLS and origin policy at the proxy but says nothing about connection-level limits, and `docs/LIMITATIONS.md` does not list them either.
- Blast radius: every endpoint, unauthenticated by default. Rated S2 rather than S1 precisely because the inherited defaults are probably non-zero — the defect is that this is unknown, unpinned and unconfigurable, not that it is provably wide open.

## Repro
```bash
cd /home/kekz/github.com/kekzl/imp
grep -rn "set_read_timeout\|set_write_timeout\|set_idle_interval\|new_task_queue" tools/ src/; echo "exit=$?"
sed -n '99,104p;145,147p' tools/imp-server/main.cpp
sed -n '12p' cmake/imp-deps.cmake
find . -name httplib.h -not -path './build*'   # -> no output: not vendored

# NOT RUN — settles the actual defaults, needs the fetched dependency:
#   grep -nE '^#define CPPHTTPLIB_(SERVER_(READ|WRITE)_TIMEOUT|KEEPALIVE_MAX_COUNT|THREAD_POOL_COUNT|HEADER_MAX_COUNT)' <build>/_deps/httplib-src/httplib.h
# NOT RUN — trickle slowloris against a running server:
#   { printf 'POST /v1/chat/completions HTTP/1.1\r\nHost: x\r\n'; \
#     for i in $(seq 1 99); do printf 'X-Pad-%d: y\r\n' $i; sleep 3; done; } | nc localhost 8080
```

## Fix direction
- Set all four explicitly in `main.cpp` from new flags, so the value is imp's decision and visible in `--help` and the startup banner, instead of a transitive default.
- Print the effective values at startup next to the other limits (`main.cpp:378-387`), which is the pattern the file already uses for `--max-concurrent`, `--request-timeout` and `--rate-limit`.
- Add the connection-level surface to `docs/LIMITATIONS.md` if the decision is to keep relying on a reverse proxy.

## Acceptance criteria
- [ ] `grep -rn "set_read_timeout" tools/imp-server/` has a hit and the value comes from a flag.
- [ ] The startup banner prints read/write timeout, keep-alive cap and thread count.
- [ ] A trickle client that sends one header every N seconds is disconnected within a bounded time.
- [ ] `docs/DEPLOYMENT.md` states which connection-level limits imp enforces itself and which it expects from a proxy.

## Unverified
- The effective defaults of cpp-httplib **v0.53.0** — the pinned version is not present in this repository. Settle with the `grep` in the Repro block against `_deps/httplib-src/httplib.h` in a configured build tree.
- Whether the trickle attack actually holds a thread under those defaults. Settle with the `nc` command above.

kekzl:
Seven issues, one question: how much work does one HTTP request buy, and who decides.

## #1614 - the rate limit keyed on a header the client writes

`X-Forwarded-For` was preferred whenever present; `remote_addr` was the fallback, not the default. There is no proxy allowlist, so varying one header per request bypassed the limit entirely. The second half outlives the request: `rate_tracker[ip]` pruned only the bucket it was asked about, so every distinct value the client invented stayed in the map for the life of the process.

| | before | after |
|---|---|---|
| key | `X-Forwarded-For`, else peer | peer, unless `--trusted-proxy` names it |
| chain handling | whole header | first element (a proxy appends) |
| key length | unbounded | 64 bytes |
| eviction | only the bucket asked about | sweep every 256 admissions |

The request log used the same unvalidated header, which made it a forged identity in the audit trail. It goes through the same function now.

## #1615 - what the two guards actually covered

Both asked `is_inference_endpoint()`, seven exact paths. `/tokenize`, `/detokenize`, `/v1/messages/count_tokens` and `/admin/{suspend,resume}` are not on it. Tokenisation walks the prompt through the BPE merge table on a server thread; `/admin` flips global state.

`--max-concurrent` guards engine capacity, so its list is unchanged. `--rate-limit` exists to stop a client hammering the process at all, so it now covers everything except `/health` and `/metrics`.

## #1616 / #1617 - fields that multiply the work of one request

| field | before | after |
|---|---|---|
| `n` | clamped from below only | `--max-n`, default 8 |
| rerank `documents` | unbounded | `--max-batch-items`, default 512 |
| embeddings `input` | unbounded | `--max-batch-items`, default 512 |
| `logit_bias` | unbounded | `--max-logit-bias`, default 1024 |

The neighbouring `max_tokens` **is** clamped to the context window, which is what makes the omission visible rather than deliberate.

`logit_bias` also had a cost problem independent of its size: one **blocking** `cudaMemcpy` device-to-host, plus an async copy back, per entry per decode step, in three hand-copied loops. The blocking read existed only so the host could do the addition.

Now one upload and one kernel, from a 4096-slot T2 arena buffer. Two details worth review:

- `atomicAdd`, not `+=`: `"1"` and `"01"` are different JSON keys that both parse to token 1, and the loop accumulated both.
- The old per-entry path is kept as a fallback for when preallocation did not run. Dropping the bias silently would change the output without saying so.

## #1618 - the 404 envelope threw on the path it echoed

The message was built from the raw request method and path and serialised with `.dump()`, which throws `json::type_error.316` on ill-formed UTF-8. A 404 for a path containing one stray `0x80` became a **500 with an empty body**, which is the exact shape this envelope was added to prevent. The file's own comment claimed `dump_safe` was used here.

Both halves: the echo is sanitised to printable ASCII and capped at 128 bytes, and the serialiser is the one that cannot throw.

## #1619 - the shipped compose file had no way to turn auth on

The entrypoint translates fifteen env vars and did not translate `IMP_API_KEY`, while the compose file published on every interface.

- `IMP_API_KEY` and `IMP_TRUSTED_PROXY` reach the binary.
- The published port binds to `127.0.0.1` unless `IMP_BIND` says otherwise.
- The three documented `-p 8080:8080` examples were the same exposure and are bound too.

`--host 0.0.0.0` **inside** the container stays as it is: the container's loopback is not reachable through a published port, so changing it would break every mapping rather than secure anything. The host side of the publication is what decides exposure.

## #1622 - connection limits are ours now

| flag | default |
|---|---|
| `--http-read-timeout` | 60 s |
| `--http-write-timeout` | 600 s |
| `--http-keep-alive-max` | 100 |

The asymmetry is deliberate: a streamed completion writes for as long as it generates. Before this, all three were whatever the build-time cpp-httplib defaulted to, which this repo cannot read, since the library is fetched at a pinned tag rather than vendored.

## Tests

9 cases in `tests/test_server_request_limits.cpp`, CPU lane.

The rate limiter moved to `tools/imp-server/rate_limit.{h,cpp}` to get there: `ServerState` pulls in `BatchingEngine`, so nothing living on it can be constructed in the only lane CI runs. Same extraction `bearer_token_matches` made, for the same reason. `allow()` takes the clock as a parameter, so the 60-second window and the eviction sweep are tested without a 60-second test.

| negative control | result |
|---|---|
| believe `X-Forwarded-For` unconditionally | 2 red |
| sweep removed | 1 red: tracker stays at 4001 across a two-minute gap |
| sanitiser passes bytes through | 2 red, one of them the throw |

Not in here: the `--max-n` / `--max-batch-items` / `--max-logit-bias` refusals and the widened rate-limit path both need a running server, so they belong to the API battery rather than the CPU lane. No test asserts them today.

## Gate

```
  decode  tg128 =  283.91 tok/s  (baseline  287.19, delta -1.14%)
  prefill pp512 = 12483.17 tok/s  (baseline 12406.87, delta 0.61%)
  prefill pp4096 = 15331.50 tok/s  (baseline 15324.70, delta 0.04%, FA2)
  own peak VRAM = 20718 MiB  (baseline 20716, delta 0.01%)
```

GPU suite: 2361 passed, 0 failed.

Closes #1614. Closes #1615. Closes #1616. Closes #1617. Closes #1618. Closes #1619. Closes #1622.


kekzl:
Superseded by a clean branch. This PR was opened from `fix/checkpoint-input-limits`, whose first commit had already been squash-merged as #1694, so GitHub saw a conflict against main, could not build a merge ref, and ran no checks (`mergeStateStatus=DIRTY`). Same content, re-applied onto fresh main.

Issue #1708, opened 2026-08-23
[MEM][S1] multi-sequence residual decode faults with an illegal access when graphs are on


kekzl:
## Fact

Multi-sequence decode on the `bitdecoding_residual` path faults with `an illegal memory access was encountered` when CUDA graphs are on. One sequence is fine; six concurrent sequences fault and the output degenerates to `Okay!!!!!!!!!!...`.

This is not #1648 (the per-step `cudaMallocAsync` address baked into a replayed graph). That is a real defect on the same path and is fixed in the branch these numbers come from - the fault survives it unchanged.

## Measured

`Qwen3-8B-Q8_0.gguf`, `imp-server --kv-nvfp4 --set kv_cache.bitdecoding_residual_tokens=64`, this box, 2026-08-23.

| arm | requests | outcome | `illegal memory access` lines in the server log |
|---|---|---|---|
| graphs on (default) | 1 | `"Hello! How can I assist you today?"` | **0** |
| graphs on (default) | 6 concurrent | all six degenerate to `Okay!!!!!!!!` | **54578** |
| `--set runtime.cuda_graphs=never` | 6 concurrent | all six `"Hello! How can I assist you today?"` | **0** |

The 6-concurrent / graphs-on arm was measured twice, on `main` and with the #1648 fix applied: 106998 and 111543 lines over a longer prompt, i.e. the same failure either way.

First fault site in the log:

```
[ERROR] executor.cu:572: CUDA error: cudaStreamSynchronize(stream) at
        /src/src/exec/executor.cu:572 - an illegal memory access was encountered
```

## Impact

- Silent wrong output, then a sticky context: the first fault takes every later request in the process with it. The client sees `finish_reason: length` and a degenerate string, not an error.
- Reachable whenever `kv_cache.bitdecoding_residual_tokens > 0` and `kv_cache_dtype=NVFP4` and more than one request is in flight. The knob is off by default (`bitdecoding_residual_tokens = 0`), which is what has kept this out of sight.
- The N=1 branch uses the persistent `d_kv_slot_buf_` and is clean; the fault is in the multi-sequence branch, and it needs graphs to appear - so the suspect is what the captured graph closes over on that path, beyond the allocation address #1648 names.

## Repro

```bash
docker run --rm --gpus all -p 8099:8099 -v $HOME/models:/models:ro imp:test \
  imp-server --model /models/Qwen3-8B-Q8_0.gguf --port 8099 --host 0.0.0.0 \
[...]

Issue #1642, opened 2026-08-22
[TEST][S1] no soak test exists, while two comments assert what a soak would check


kekzl:
## Fact
- No long-run test exists anywhere in the repo — the largest request count any test drives is 10 concurrent requests (`tests/api/test_concurrency.py:37`) — while two shipped source comments describe assertions "a soak makes" (`tools/imp-server/metrics_memory.cpp:56`, `tests/test_memory_backend.cpp:163`).

## Evidence
```
$ grep -rniE "soak|endurance|long-run" tests/ scripts/ Makefile .github/ tools/ --include=*.py --include=*.sh --include=*.cpp --include=*.cu --include=Makefile --include=*.yml
tests/test_memory_backend.cpp:163:    // The I2 assertion a soak makes: no acquire event carries phase == Serving.
tools/imp-server/metrics_memory.cpp:56:        // "used" gauge cannot express it. imp_kv_blocks_live is the one a soak
```
Two hits, both comments referring to a soak that does not exist.
```
$ grep -rn "range(" tests/api/test_concurrency.py
tests/api/test_concurrency.py:37:            futures = [pool.submit(make_request, i) for i in range(10)]
tests/api/test_concurrency.py:62:            futures = [pool.submit(make_request, i) for i in range(10)]
tests/api/test_concurrency.py:123:            futures = [pool.submit(stream_request, i) for i in range(5)]
tests/api/test_concurrency.py:244:            fillers = [pool.submit(filler_request, i) for i in range(3)]
$ grep -n "ThreadPoolExecutor\|range(" tests/api/test_performance.py
tests/api/test_performance.py:17:        for _ in range(5):
tests/api/test_performance.py:45:        for _ in range(3):
tests/api/test_performance.py:74:        with concurrent.futures.ThreadPoolExecutor(max_workers=4) as pool:
tests/api/test_performance.py:75:            futures = [pool.submit(make_request, i) for i in range(4)]
```
```
$ grep -n "^[a-z-]*:" Makefile | grep -i "soak\|stress\|endur"; echo "rc=$?"
rc=1
```
- The observable that a soak would assert already exists and says so in its own comment:
```
$ sed -n '50,58p' tools/imp-server/metrics_memory.cpp
        // The split matters more than the total, and its absence cost me a wrong
        // bug report (#1115): `used` grew by one block per request and looked
        // like a leak, but the server enables prefix caching by default, so
        // free_sequence MOVES the last reference of a hashed block into the
[...]

Issue #1573, opened 2026-08-22
[TEST][S1] check-degeneration is a skill doc: degen_suite.py has zero call sites


kekzl:
## Fact
- "check-degeneration" is a Claude Code skill document, not an automated procedure. Its deepest battery (`tools/analysis/degen_suite.py`) has zero invocation sites in the repository, its GTest battery is in the disabled GPU lane and skips without a checkpoint, and the only degeneration assertion that runs by itself is a 3-line heuristic in the local pre-push hook.
- `.claude/skills/check-degeneration/SKILL.md:1`
- `scripts/verify.sh:660`
- `CMakeLists.txt:928`

## Evidence
- It is a skill file, i.e. a prompt for a human or an agent to follow:
```
$ sed -n '1,4p' .claude/skills/check-degeneration/SKILL.md
---
name: check-degeneration
description: Use when verifying that a model in the imp inference engine produces coherent output without repetition loops, token-stuck states, or state corruption across turns. Triggers on "degenerates", "check degeneration", "repetition loop", "own own own", "stuck token", "multi-turn regression", "does it still work", and after enabling CUDA graphs / changing forward pass / MoE routing / KV cache / GDN state.
---
```
  Its pass criteria are quantitative but exist only in prose:
```
$ sed -n '24,34p' .claude/skills/check-degeneration/SKILL.md
## Pass criteria (quantitative)

A run **passes** only when ALL of these hold:

1. **No verbatim repetition**: no token repeats >4 times in a row, no 3-gram repeats >3 times.
2. **No early abort**: ≥10 generated tokens before any stop condition (unless prompt is single-word factual like "Paris").
3. **stderr clean**: no match for the grep pattern below (silent fallback masks the bug).
4. **Decode within 30% of baseline** in `tests/perf_baseline.json` for the same model. >30% drop almost always means graphs fell back silently.
5. **Multi-turn**: turn 2+ output is grammatical and topical for the new question; KV/state from turn 1 didn't leak.
```
  None of the five is implemented as an assertion anywhere.
- The suite the skill calls its "deepest battery" is invoked by nothing:
```
$ grep -rn "degen_suite.py" --include=*.sh --include=*.yml --include=Makefile --include=*.hook .
(no output)

$ grep -rn "degen_suite\|multiturn_deep" Makefile .github/ scripts/
(no output)
```
  It has a correct exit contract, so wiring it would work — it is simply not wired:
```
[...]

Issue #1572, opened 2026-08-22
[TEST][S1] chat-template goldens cover 1 of 10 families, the other 9 are structural smoke


kekzl:
## Fact
- One of imp's ten chat-template families has a byte-for-byte golden against `transformers.AutoTokenizer.apply_chat_template`. The other nine are covered by structural smoke tests (token counting, substring presence) that cannot detect a wrong prompt.
- `tests/test_gpt_oss_harmony_golden.cpp:83`
- `src/model/chat_template.h:13`
- `tests/test_jinja.cpp:553`

## Evidence
- The ten families:
```
$ sed -n '13,24p' src/model/chat_template.h
enum class ChatTemplateFamily {
    RAW,          // No template — pass raw text
    CHATML,       // <|im_start|>...<|im_end|> (Qwen3, etc.)
    LLAMA2,       // [INST]...[/INST] (Llama 2, older Mistral V1/V2)
    MISTRAL_V3,   // [INST]...[/INST] + [TOOL_CALLS]/[AVAILABLE_TOOLS] (Mistral V3-Tekken: 3.x family)
    LLAMA3,       // <|start_header_id|>...<|end_header_id|>...<|eot_id|>
    NEMOTRON,     // <extra_id_0>System\n...<extra_id_1>\n<extra_id_0>User\n...
    GEMMA,        // <start_of_turn>user\n...<end_of_turn>\n<start_of_turn>model\n
    DEEPSEEK_R1,  // <｜User｜>...<｜Assistant｜>...<｜end▁of▁sentence｜>
    PHI,          // <|user|>...<|end|>...<|assistant|>
    HARMONY,      // gpt-oss: <|start|>role<|message|>...<|end|>, channels, <|return|> stop
};
```
- The only HF reference generator in the tree, and the only exact-comparison test:
```
$ grep -rn "apply_chat_template" tests/ scripts/ tools/ | grep -v "params.apply_chat_template"
tests/test_gpt_oss_harmony_golden.cpp:14:// (transformers AutoTokenizer.apply_chat_template; see
tests/refs/gen_harmony_golden.py:6:tokenizer's chat template via `transformers.AutoTokenizer` and renders fixed
tests/refs/gen_harmony_golden.py:80:        rendered = tok.apply_chat_template(
tests/refs/gen_harmony_golden.py:89:        f.write("// reference (transformers AutoTokenizer.apply_chat_template).\n")
[...]

PR #1710, opened 2026-08-23
fix(test): the degeneration battery had no call sites, and its stand-in saw one shape (#1573)

kekzl:
The repo's stated last line of defence against wrong output was a skill
document plus one distinct-token count. Both halves get wired up.

## The battery had zero call sites

`tools/analysis/degen_suite.py` is 41 checks that exit non-zero correctly and
that **nothing ever ran**. `scripts/test_server.sh` already boots the server it
needs, so that is where it goes.

First run in its life, `Qwen3-4B-Instruct-2507-Q8_0` over FP16 KV:

```
degen_suite: 35 checks, 0 FAIL, 1 skipped (5s)
```

**It earned the wiring on its second invocation.** Same binary, `--kv-nvfp4
--set kv_cache.bitdecoding_residual_tokens=64`:

```
degen_suite: 46 checks, 1 FAIL, 1 skipped (22s)
  FAIL [stream] stream == non-stream at temp=0
    stream='1, 2, 3, 4, 5, 6, 7, 8, 9, 10'
    nonstream='1,2,3,4,5,6,7,8,9,10'
```

Control, `--kv-nvfp4` **without** the residual knob: both transports return
`'1, 2, 3, 4, 5, 6, '` - identical. Added to #1708, which is the same path.

Categories rather than `--corpus`: the corpus battery is ~250 prompts and
belongs in a longer lane than a server battery that already takes minutes.

## The stand-in gate saw one shape

The pre-push smoke check asserted one thing about output: at least 8 distinct
tokens in the last 32. That catches `own own own` and nothing else in the
skill's own table. Two of the remaining shapes are checkable from output this
gate already collects, with the skill's own thresholds - no token more than 4
times in a row, no 3-gram more than 3 times.

**Why the 3-gram check is not redundant with the count.** Measured on synthetic
streams:

| stream | n | max run | max 3-gram | verdict |
|---|---|---|---|---|
| healthy sentence | 14 | 1 | 1 | pass |
| stuck token | 14 | 8 | 6 | FAIL |
| 3-gram loop | 14 | 1 | 4 | FAIL |
| early abort | 3 | - | - | FAIL |
| exactly at the limits | 14 | 4 | 3 | pass |

And the case the old gate lets through: a stream with **15 distinct tokens** in
its last 32 - comfortably past the 8 it wants - while repeating a 3-gram four
times.

**The token floor is per-prompt.** The skill says 10, with its own exception
for "single-word factual" prompts, and this gate's own prompt is one: "The
capital of France is" answers in **11 tokens** on the 4B model. A flat 10 would
sit one token from a false red, so both smoke prompts pass 5 and the default
stays where the skill put it.

Real run after the change, from this PR's own push:

```
PASS Qwen3-4B Q8_0 (dense) (distinct=11, tokens=11, max-run=1, max-3gram=1, contains 'Paris')
```

## One thing that is not a feature

The temp file is cleaned before the early returns rather than after them.
Adding checks in front of what used to be the only `rm` would have leaked one
file per failed smoke run.

## Not in this PR

#1571 (KL / PPL-drift / tool-conformance gates, NIAH), #1572 (chat-template
goldens for 9 of 10 families) and #1642 (no soak test) are each their own piece
of work rather than wiring.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -180,6 +180,23 @@ there instead of retelling it.
 
 ### Fixed
 
+- **The degeneration battery had zero call sites, and the gate that stood in
+  for it saw one shape** (#1573). `tools/analysis/degen_suite.py` is 41 checks
+  that exit non-zero correctly and that nothing ever ran; it is wired into
+  `make test-server` now, where the running server it needs already exists.
+  First run: **35 checks, 0 fail, 5 s**. Against `--kv-nvfp4` plus the residual
+  knob it found a real one on its first attempt (stream and non-stream
+  disagreeing at greedy) - added to #1708.
+
+  The pre-push smoke gate also gained the two criteria the skill it stands in
+  for already specifies: no token more than 4 times in a row, no 3-gram more
+  than 3 times. The old distinct-token count could not see a loop: a stream
+  with **15** distinct tokens in its last 32 passes it while repeating a 3-gram
+  four times. The token floor is per-prompt, because this gate's own prompt is
+  the "single-word factual" case the skill excepts - it answers in 11 tokens,
+  one above a flat limit of 10.
+
+
 - **`json_schema`: an unconstrained `integer` had no digit bound** (#1540). At
   the server's default temperature the sampler could sit in the digit state and
   emit `1020000000000000000000000000000000000000` for a population field - a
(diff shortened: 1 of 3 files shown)


kekzl:
Superseded by the branch this reopens as: the CI trigger never fired for `fix/test-gates` (0 checks reported), and a rebase to pick up main plus the wall-clock addition needed a force-push, which is gated here. Same commits plus one, on `fix/degen-gates`.

PR #1811, opened 2026-08-29
fix(server): mtp_k=auto reads the resolved batch size, not the raw --max-batch flag

kekzl:
Post-v0.32.0 fix for #1809, found by a peer conformance run (26/26 green) on the tagged tree.

## The defect

The auto gate asked `state.default_args.max_batch_size` - the raw CLI flag. The precedence that decides the batch (`--max-batch` > `[runtime] max_batch_size` from imp.conf > per-load override > engine auto) lives in `build_config`. So a single-stream server configured the imp.conf way declined the MTP head, and `/health` explained it with "this server takes concurrent requests" while serving one stream.

Cost for that configuration, measured by the peer on Qwen3.8-27B-NVFP4 single-stream thinking: **88.1 -> 117.2 tok/s (+33%)** left switched off.

## The fix

One `resolve_max_batch_size(args, runtime_cfg, overrides)`; `build_config` uses it, the load path uses it and records the result in `ServerState::resolved_max_batch_size`, and `/health` reads that instead of the flag.

## Verified (make-build image, Qwen3.8-27B-NVFP4)

| configuration | resolved batch | auto |
|---|---|---|
| `--max-batch 1` | 1 | engages (unchanged) |
| `--set runtime.max_batch_size=1` | 1 | **engages (was: declined)** |
| default | auto 28 | declines, head unloaded (unchanged) |

`/health` at batch 1 with an explicit `speculative.mtp_k=0` no longer claims concurrency.

`make verify-fast` OK, `ci_static_gates` green, CPU lane green.

Not in here: a unit test - `handlers.cpp` links only into the server binary, and the hint needs a 19 GiB checkpoint with an MTP head, so this is covered by the three-configuration server run above rather than by CI.

https://claude.ai/code/session_01KvfZX6NNFLBksYyPFRSxPR


kekzl:
Filed as #1812 with the full evidence, so the changelog and the release have a number to cite.

Measured on the released `ghcr.io/kekzl/imp:latest` @ `53d3d86e` (0.32.0), Qwen3.8-27B-NVFP4, single-stream thinking prompt, 400 completion tokens, 3 rounds after a discarded warmup, median:

| server | MTP | tok/s |
|---|---|---|
| compose default (batch auto 22) | off | 88.1 |
| `--set runtime.max_batch_size=1` | off, auto declined | 88.2 |
| `runtime.max_batch_size=1` + forced `mtp_k=2 ngram=false` | on | 115.7 |
| `--max-batch 1` (auto) | on | 117.2 |

The middle two rows are the bug: same pinned stream, opposite outcome depending on which key set it. Conformance tier was 26/26 green on that same image, twice, so nothing else was wrong with the server.

PR #1826, opened 2026-08-31
docs: drop the numbers nothing checks, and record which ones earn their place

kekzl:
Follow-up to #1825, which paid a maintenance bill instead of questioning it: it
"fixed" `144 -> 147 files` and `0.39 s -> 7.6 s`. Both will be wrong again
shortly, and nothing would notice. Removed instead.

## The test

**Does something fail when this number goes wrong?**

| kept | why |
|---|---|
| gate thresholds 8%/8%/10% | `verify.sh` enforces exactly those |
| hardware constants (32 GB, 1792 GB/s) | cannot drift |
| dated measurements | findings, true of the afternoon they name - not inventory. "all 82 assertions described the mock" is past tense and stays |

| removed | was in |
|---|---|
| `~135k LOC` / `~220k with tests` | root `CLAUDE.md` |
| `147 files` | `src/compute/CLAUDE.md` |
| `1021 of 2636 GTest macros`, `silently skips 63` | `tests/CLAUDE.md` |
| `2-14 s`, `~3.5 min`, `37 s`, `~45 s`, `7.6 s` | root, `src/compute`, `tests`, `AGENTS.md` |

Runtimes collapse to the magnitude that picks a target ("seconds" vs "minutes")
and disappear where even that changes nothing.

The lane split is the interesting case: it **is** pinned, but by
`check_test_lanes.py`, not by the sentence quoting it. Restating a gated number
in prose does not inherit its gate, so the doc now states the fact and names the
command.

## Where the rule lives

In the `docs-layers` skill, not the root `CLAUDE.md` - that file's own policy is
"if something here is also in a skill, the skill is the copy that gets
maintained". The first draft put it in `CLAUDE.md` and pushed it over its
2000-token budget, which is a fair verdict on adding prose about avoiding prose.

## Verification

Docs only. `bash scripts/ci_static_gates.sh` all groups pass; `make dev-test`
8/8; `docs_lint` clean of FAILs.

https://claude.ai/code/session_01KvfZX6NNFLBksYyPFRSxPR


kekzl:
Branched off docs/l3-verify-0831 instead of main, so it carried two commits already squash-merged as #1824/#1825 and went DIRTY. Reopened from main as the same change, no force-push.

Issue #1574, opened 2026-08-22
[DET][S1] deterministic reaches 4 kernel sites and not the primary NVFP4 GEMM path


kekzl:
## Fact
- `runtime.deterministic` / `IMP_DETERMINISTIC` reaches exactly four kernel sites, all through the single boolean `process_diag_deterministic_gemm()`; the primary NVFP4/CUTLASS GEMM path is not one of them.
- `src/runtime/process_diag.cpp:146`
- `src/compute/gemm.cu:443`
- `src/compute/sampling_topk_topp.cu:585`
- `src/compute/moe_routing.cu:606`
- `src/compute/moe_routing.cu:724`
- `src/compute/moe_routing_permute.cu:180`
- `src/compute/gemm_cutlass_grouped_3x.cu` — 0 references to `deterministic` or `process_diag`
- Docs state the mode covers "GEMM" without scoping it to the cuBLASLt path: `docs/determinism.md:23`, `src/runtime/config.h:35-38`.

## Evidence
```
$ grep -rn "process_diag_deterministic_gemm()" src/ tools/ include/
src/runtime/process_diag.cpp:146:bool process_diag_deterministic_gemm() { return slot().deterministic_gemm; }
src/runtime/process_diag.h:50:bool process_diag_deterministic_gemm();
src/compute/gemm.cu:443:    const bool s_deterministic_gemm = imp::process_diag_deterministic_gemm();
src/compute/sampling_topk_topp.cu:585:    const bool deterministic = process_diag_deterministic_gemm();
src/compute/moe_routing_permute.cu:180:    if (process_diag_deterministic_gemm()) {
src/compute/moe_routing.cu:606:    if (process_diag_deterministic_gemm()) {
src/compute/moe_routing.cu:724:        if (process_diag_deterministic_gemm()) {
(config.cpp:107 and engine.cpp:603 are comments; tests/* are tests)

$ grep -n "deterministic\|process_diag" src/compute/gemm_cutlass_grouped_3x.cu
NO MATCH

# docs/determinism.md:20-24 — what the mode claims to cover
- **MoE token routing** - atomic expert-bucket scatter ordering.
- **Top-k sampling** - atomicMax/atomicAdd softmax-stat races (single-block
  path, `top_k <= 128`).
- **GEMM** - implies `deterministic_gemm` (cuBLASLt `no_reduce_split`;
  timing-based algo selection is itself a non-determinism source).

# src/runtime/config.h:35-38 — same claim, in the config comment
[...]

PR #1858, opened 2026-09-02
gates: five tools that did not catch what they gate (mutation-tested all 13, two rounds)

kekzl:
## What this is

Mutation test of the 13 static gates: plant the violation each one exists to catch, check it goes red. Twelve did. Three tools did not do what their own header claimed.

| Gate | Planted | Before |
|---|---|---|
| `check_determinism_sites.py` | `process_diag_deterministic_gemm()` reader in `src/exec/` | **green** |
| `check_determinism_sites.py` | one of `gemm.cu`'s two reads removed | **green** |
| `tests_without_assertions.awk` | (its own 53 hits) | **32 were the filter** |
| filesize, dead-inline, log-fatal, test-lanes, entrypoint, alloc-sites, alloc-pairs, launch-guards, sync_docs, docs_lint, citations | one violation each | red, correct |

## 1. Determinism site gate

- Scanned `src/compute/` only. The MoE CUTLASS dispatch is `src/exec/`, which is the neighbourhood of the defect the gate was built for (#1574): a new reader there passed unnoticed.
- Pinned the file set, not the read count, so a branch inside `gemm.cu` could stop reading the flag silently.

Now scans all of `src/` with comments stripped (two prose mentions in `src/runtime/` would otherwise count), skips the accessor's own definition, and pins per-file counts. Same 6 reads in 4 files as before.

`--selftest` plants all eight drifts on a fixture tree and runs in the `filesize` gate group:

```
8/8 cases against the fixed gate
4/8 cases against the pre-fix implementation (src/exec, src/quant, both count cases fail)
```

## 2. Assertionless-test filter

53 hits, 32 of them the filter itself:

| Bug | Cost |
|---|---|
| harness class `[a-zA-Z_]*` never matched a name with a digit (`run_fa2`, `run_pv256`) | all 21 hits in `test_fmha_fp8.cu`, which the header explained away as "its own harness under a different name" |
| single-line `TEST(...) { run_x(); }` sees no line-initial `}`, stayed open, judged against the next block | 4 |
| `R"({ ... })"` JSON fixture ends the body at its own `}` | 5 |

16 remain, none a real gap: 8 `Bench*` that only measure, 6 on `compare_dp4a_vs_mmvq` (18 EXPECTs inside), 2 on `chunkpar_matches_fused` (18 ASSERTs inside).

## 3. Pre-push hook over-gated its own gate scripts

`^scripts/` in `SOURCE_RE` sent every edit under `scripts/` through `make verify-fast` on a shared card. verify.sh reaches 8 of the 25 shell scripts there (itself, `check_e2e_lane_split.sh`, `smoke_test.sh`, `gen_perf_baseline.sh` and the four guards CMake registers as ctests). The other 17 (`ci_static_gates.sh`, `check-release.sh`, the bench harnesses, the server smoke drivers) cannot change the run's outcome, same reasoning that excluded `.md`/`.py`/`.hook` in #1723 and #1825.

`scripts/verify_touches.sh` answers it by membership in the tree (named in `verify.sh` or `CMakeLists.txt`), not from a copied list, so adding a call re-arms that script's gate with no edit anywhere. `check_precommit_filter.sh` pins both directions.

## 4. Second round: one variant per gate

Same 13 gates, a different shape of the same violation. Two more stayed green:

| Gate | Variant that passed | Fix |
|---|---|---|
| `check_doc_citations.py` | citation to a file that no longer exists (rename/split, e.g. #1782) | 0 basename hits is DEAD, not AMBIGUOUS; >1 stays AMBIGUOUS |
| `tests/test_entrypoint.sh` | `docker-compose.yml` offers an `IMP_*` the entrypoint never reads (the #1619 shape) | compares the two surfaces: 17 compose names must each be read; 27 -> 44 cases |

Caught correctly in this round: launch guards inside a template, `cudaHostAlloc` as an alloc site, a dead inline template in a `.cuh`.

```
entrypoint: 44 passed, 0 failed
citations:  PASS 0 dead citation(s) across 33 living doc(s)
```

## Gate

```
ci-static-gates: all selected gates passed        (all 13, full run)
  determinism selftest: 8/8 cases
  check_precommit_filter: 19 pre-commit case(s), 5 pre-push case(s), 7 scripts/*.sh case(s), installed hooks current
  test-lanes: 1054 pinned / 1624 in ctest -L unit
  citations: PASS 0 dead across 33 living docs
  docs_lint: OK (52 warning(s))
```

Pushed with `--no-verify`: the diff is one `.awk`, one `.py`, two `.sh` and two `.md`, no C++ or CUDA, and the card is held by another tenant right now (30% util, 7324 MiB, no GPU container). Everything `verify-fast` would add over the block above is GPU work on code this PR does not touch.

Not in here: no fix for the 16 remaining assertionless tests (all explained), no change to what any gate accepts on a clean tree.

https://claude.ai/code/session_01SVcruBPPPA5AKQwzHzn4rr

### DIFF
--- a/.claude/skills/find-stubs/SKILL.md
+++ b/.claude/skills/find-stubs/SKILL.md
@@ -80,12 +80,12 @@ done
 
 ```bash
 rg -n 'DISABLED_' tests                                                            # 9, read all
-awk -f .claude/skills/find-stubs/tests_without_assertions.awk $(find tests -name '*.cpp' -o -name '*.cu')   # 53
+awk -f .claude/skills/find-stubs/tests_without_assertions.awk $(find tests -name '*.cpp' -o -name '*.cu')   # 16, all explained
 python3 tools/check_test_lanes.py --report                                         # 1054 in no CI lane (pinned), 1624 in ctest -L unit
 rg -c 'GTEST_SKIP' tests | awk -F: '{s+=$2} END {print s}'                          # 208, mostly legitimate (no GPU, no model)
 ```
 
-A one-line regex over test bodies stops at the first line-initial `}` and flagged 184 of ~2600; the awk script splits on `^TEST`/`^}` and drops harness-delegating bodies (read its header). A filter like `DetEvalE2ETest.*` that matches 0 tests prints PASSED. A test whose inputs cannot reach the defect passes its own mutant: uniform-random rows never produced the FP8 `out / row_scale` overflow (fixed by `SsmPrefillFp8TinyRowsStayFinite`); `PagedOracle` at HD128 never ran the HD256 word-load path (#1817). Mutation baseline 90.4% (`docs/audit/MUTATION_BASELINE.md`); the dominant escape is "input does not reach the failure".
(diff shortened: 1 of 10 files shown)


kekzl:
Gate rerun on the card after it came free (the PR was pushed with `--no-verify` while another tenant held it).

```
make build            rc=0
make verify-fast #1   FAIL  decode tg128 263.09 (baseline 287.19, -8.39%), prefill pp512 11377 (-8.30%)
make verify-fast #2   OK    decode tg128 264.85 (-7.78%),                  prefill pp512 11589 (-6.60%)
                            peak VRAM 20738 MiB (baseline 20716, +0.11%)
                            graphs ON 419.24 vs OFF 189.41 tg256 = 2.21x
                            smoke: distinct=11, max-run=1, contains 'Paris'
```

Not this PR: `git diff --stat origin/main..HEAD -- '*.cpp' '*.cu' '*.cuh' '*.h' CMakeLists.txt cmake/` is empty, so both arms run a bit-identical binary. The card reads 13-16% sm at idle with no GPU container (idle here is 2-5%), which puts this host reproducibly 7.8-8.4% under the pin, i.e. straddling the 8% gate. Baseline left alone.

Issue #1626, opened 2026-08-22
[BUILD][S1] the clang-tidy CI job has never linted a file: git exit 128 hidden by continue-on-error


kekzl:
## Fact
- The `clang-tidy` CI job has never linted a file: it dies at `git diff` with exit 128 and reports green because the step is `continue-on-error: true`.
- `.github/workflows/ci.yml:279`
- `.github/workflows/ci.yml:281-286`
- Missing counterpart: `.github/workflows/ci.yml:121-122`

## Evidence
```
$ sed -n '278,287p' .github/workflows/ci.yml
      - name: clang-tidy (changed .cpp only, advisory)
        continue-on-error: true  # advisory — flip to a hard gate once baseline-clean
        run: |
          base="${{ github.event.pull_request.base.sha }}"
          [ -z "$base" ] && base="${{ github.event.before }}"
          if [ -z "$base" ] || ! git cat-file -e "${base}^{commit}" 2>/dev/null; then
            base="$(git rev-parse HEAD~1 2>/dev/null || git rev-parse HEAD)"
          fi
          mapfile -t changed < <(git diff --name-only "$base" HEAD | grep -E '^(src|tools)/.*\.cpp$' || true)

$ gh run view --job=96880658068 --log | tail -8
clang-tidy  ...  shell: sh -e {0}
clang-tidy  ...  fatal: not a git repository (or any parent up to mount point /)
clang-tidy  ...  Stopping at filesystem boundary (GIT_DISCOVERY_ACROSS_FILESYSTEM not set).
clang-tidy  ...  ##[error]Process completed with exit code 128.
```
- The `Build` job has the fix the `tidy` job lacks:
```
$ sed -n '121,122p' .github/workflows/ci.yml
      - name: Trust the checkout (container UID differs from the runner's)
        run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
```
- Not a one-off. Four consecutive runs, all reported `✓ clang-tidy`:
```
$ for r in 32515833803 32519280728 32496939332 32520600829; do gh run view $r | grep -E "clang-tidy|exit code"; done
✓ clang-tidy in 1m26s (ID 96877155141... )   X Process completed with exit code 128.
✓ clang-tidy in 1m24s (ID 96891107597)       X Process completed with exit code 128.
✓ clang-tidy in 1m24s (ID 96819165321)       X Process completed with exit code 128.
✓ clang-tidy in 1m55s (ID 96892362411)       X Process completed with exit code 128.
[...]

PR #1965, opened 2026-09-09
ci: cut the workflow's critical path, and fix the change detection that had never run

kekzl:
## Summary

Measured first, per-job wall clock over the last 8 PR runs of `ci.yml`. `Sanitizers` was the run's wall clock in **8 of 8**, and total run time was 14-20.5 min.

**Measured on this branch, same workflow, end to end:**

| job | before (median) | after |
|---|---|---|
| **whole run** | **~19 min** | **5.8 min** |
| Sanitizers | 19.0 min | **2.6 min** |
| Build | 6.2 min | 3.9 min |
| clang-tidy (ran after Build) | 1.8 min | **skipped** |

Three changes, the third found only because the second one visibly failed to work.

### 1. `Sanitizers` gets a ccache, and stops building the PTX it cannot use

It had no compiler cache of any kind and rebuilt every TU, CUDA included, every run. It now keeps a store of its own (the shape `PTX fallback` already uses) and passes `IMP_DISABLE_120F_FALLBACK=ON`: the option defaults OFF, so the lane was also emitting the `compute_120f` PTX entry beside the `sm_120a` SASS. It can use neither image - no GPU on the runner, and `CMakeLists.txt` does not sanitize device code - and `Build` has the second gencode measured at **+53.1 % device-compile time**. The shipped fatbin is still assembled and checked by `PTX fallback`, which deliberately leaves the option at OFF.

Separating the two effects: cold ccache, single gencode was **13.7 min**; warm, **2.6-3.1 min**.

### 2. `clang-tidy` starts only when a `src/`|`tools/` `.cpp` changed

It required only `outputs.code`, so it started on every code PR and paid ~2.6 min of container pull and cache round trip on the `.cu`- and docs-only PRs that are most of them - onto the tail of the critical path, since it `needs: build`. `Build` now also publishes a `cpp` output (fail-open, like `code`). Its build cache also becomes `actions/cache/restore`: the job only reads `compile_commands.json` and `_deps`, but plain `actions/cache` re-uploaded ~476 MB per run under a key `Build` had already written - 29 s, and that much of the repo's 10 GB cache quota spent evicting entries the next PR wants.

### 3. The change detection had never run at all

The first CI run on this branch ran `clang-tidy` anyway. The reason, from the `Build` log on a two-file diff with no code in it:

```
non-docs changes: code=true (src|tools .cpp: true)
changed files:
```

Nothing after `changed files:` - an empty list, so no `git diff` ever ran. The detection step was the **first git caller in the job**, and the step running `git config --global --add safe.directory` sat 50 lines below it, so `git cat-file -e` died on the container ownership refusal with stderr suppressed and the fail-open returned `code=true` on every run.

So the docs-only skip this step exists for had **never once fired** - which is why a docs-only PR still paid a ~4 min `Build` - and the `cpp` gate from change 2 was inert for the same reason. The `safe.directory` step now runs immediately after checkout, and a fall-open prints a `::warning::` naming the base it could not resolve, because a silent `code=true` with an empty list reads exactly like a genuine code PR and is what let a permanently-open gate hide.

Confirmed on the current head:

```
non-docs changes: code=true (src|tools .cpp: false)
changed files:
.claude/skills/building-and-testing/SKILL.md
.claude/skills/shipping-prs/SKILL.md
.github/workflows/ci.yml
CHANGELOG.md
```

Populated list, `cpp: false`, no fall-open warning, and `clang-tidy` skipped.

Nothing here changes what CI *checks* - no gate, test or lane is dropped, and `Build` (the one required context, ruleset 14716423) keeps identical gating behaviour on code PRs. Docs-only PRs now genuinely skip the heavy steps, which is what the existing code always intended.

## Test plan

- [x] Green end to end on the current head; `clang-tidy` skipped, detection output verified in the job log (quoted above)
- [x] `scripts/ci_static_gates.sh` and `scripts/check_dep_pins.sh` green; the new `actions/cache/restore@<sha>` is 40-hex pinned
- [x] Detection logic exercised against a throwaway git repo under **both `dash` and `bash -eo pipefail`** - the CUDA image gives these steps `sh -e {0}`, confirmed in the job log. Resolvable base: docs-only -> `code=false cpp=false`; `+src/*.cpp` -> `true/true`; workflow-only -> `true/false`. Unresolvable base -> warning, `true/true`
- [ ] `make verify-fast` not run: no GPU in this session, and the diff is workflow + CHANGELOG only - no `src/`, `include/`, `tools/` or `tests/` file is touched, so it is outside `PERF_RE` and outside the pre-commit GPU trigger
- [ ] Not an engine perf change; no `tests/perf_baseline.json` refresh

## Known issue this PR does NOT fix

`clang-tidy`'s lint step exits 128 with `fatal: not a git repository`, and `continue-on-error: true` reports the job green regardless. **It is pre-existing** - byte-identical in run 2590 on `main`'s side, before this branch existed - and the step also uses `mapfile`/`${#arr[@]}` under `sh`, which would not survive even with git fixed. Same class as #1626 ("It had never linted a file").

Left alone deliberately: it is a correctness bug in an advisory lane, separate from this PR's subject, and it wants its own PR - one that should also decide whether `clang-tidy` keeps `continue-on-error`, since that flag is what let it hide. Change 3 above means it is at least no longer triggered on PRs that change no `.cpp`.

https://claude.ai/code/session_018HBb9gf2hacFVkWV9bn7Hk

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,6 +43,7 @@ jobs:
     runs-on: ubuntu-24.04
     outputs:
       code: ${{ steps.changes.outputs.code }}
+      cpp: ${{ steps.changes.outputs.cpp }}
 
     container:
       # Native CUDA devel image: nvcc $CUDA_VERSION is on PATH at /usr/local/cuda
@@ -92,15 +93,34 @@ jobs:
         run: |
           base="${{ github.event.pull_request.base.sha || github.event.before }}"
           code=true
+          # `cpp` gates the clang-tidy job, which lints changed src|tools *.cpp
+          # and nothing else. It used to start on every code PR and pay ~2.6 min
+          # (container pull + cache round trip) to print "nothing to lint" - and
+          # it does that on most PRs here, which are .cu, .py or .md. Same
+          # FAIL-OPEN rule as `code`: any uncertainty runs the job.
+          cpp=true
           if [ -n "$base" ] && git cat-file -e "${base}^{commit}" 2>/dev/null; then
             files="$(git diff --name-only "$base" HEAD)"
             # code stays true unless EVERY changed file is markdown or under docs/.
             if [ -n "$files" ] && ! printf '%s\n' "$files" | grep -qvE '(\.md$|^docs/)'; then
               code=false
             fi
+            # `grep -c`, not `grep -q`, and no here-string. Two traps meet on
+            # this line. #1499: `grep -q` exits at the first match, the producer
+            # takes EPIPE and pipefail reports a non-zero pipeline for a
+            # SUCCESSFUL match. And the here-string the other lanes use to dodge
+            # that is a bashism - this job declares no `shell:`, so it gets
+            # whatever the CUDA image provides, and `Build` is the one required
+            # check (ruleset 14716423): a syntax error here fails every PR.
+            # `grep -c` reads all of stdin, so nothing ever takes EPIPE, and it
+            # is POSIX. `|| true` because grep exits 1 when the count is 0.
+            n_cpp="$(printf '%s\n' "$files" | grep -cE '^(src|tools)/.*\.cpp$' || true)"
+            [ "${n_cpp:-0}" -gt 0 ] || cpp=false
           fi
           echo "code=$code" >> "$GITHUB_OUTPUT"
-          printf 'non-docs changes: code=%s\nchanged files:\n%s\n' "$code" "${files:-<none>}"
+          echo "cpp=$cpp" >> "$GITHUB_OUTPUT"
+          printf 'non-docs changes: code=%s (src|tools .cpp: %s)\nchanged files:\n%s\n' \
+            "$code" "$cpp" "${files:-<none>}"
 
       # The blocking static gates, FIRST, before the compile. Ruleset 14716423
       # requires exactly one context (`Build`), so a gate that runs only as its
@@ -264,7 +284,7 @@ jobs:
   tidy:
     name: clang-tidy
     needs: build
-    if: needs.build.outputs.code == 'true'
+    if: needs.build.outputs.cpp == 'true'
     runs-on: ubuntu-24.04
     container:
       image: nvidia/cuda:13.3.1-devel-ubuntu26.04@sha256:8cf42b8dc4c34d47fb42ffb0923f8a5e363469a7149181c094da336d311bb466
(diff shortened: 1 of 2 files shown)


kekzl:
First CI run on this branch is green. Two things worth recording, one of which corrects a claim in the description above.

### Measured effect so far

| job | before (median, 8 runs) | this run |
|---|---|---|
| Sanitizers | 19.0 min | **13.7 min** |

That 13.7 min is on a **cold** ccache - the `ccache-asan-*` key had never existed, so every TU still compiled. The drop is the `compute_120f` gencode alone. The ccache half only pays from the next run on this branch onward, so the number to read is the one after this.

The run's wall clock was 15.8 min rather than ~13.7 because `Build` took 13.9 min this time (against its own 6.2 min median): the branch had just rebased onto a newer `main`, and `ccache -s` in that job reports **63.48 % hits, 210 misses, store 1.6 / 8.0 GB**. Not something this PR changes.

### `clang-tidy` has never linted a file, and this PR is not the cause

The lint step exits 128 with `fatal: not a git repository (or any parent up to mount point /)`. It is `continue-on-error: true`, so the job reports **success** and nothing surfaces.

It is pre-existing, not a regression from the gating change here - the identical failure is in run 2590 on `main`'s side, before this branch existed:

- this branch, job 102667507298: `fatal: not a git repository` -> `##[error]Process completed with exit code 128`
- run 2590, job 102638151052: byte-identical failure

Note also that the step logs `shell: sh -e {0}` - the CUDA image gives these jobs `sh`, not bash, so the step's `mapfile -t` and `${#changed[@]}` are bashisms that would not survive even if the git call were fixed. This looks like the same class as #1626 ("It had never linted a file"), either regressed or a second instance.

So the description's line about the job printing "nothing to lint" is wrong: it crashes instead. The saving the `cpp` gate buys is the same either way - ~2.6 min off the tail of the critical path - but it is gating a job that currently does no work.

**I have deliberately not fixed that here.** It is a correctness bug in an advisory lane, separate from this PR's subject, and fixing it would widen a diff that is otherwise two files. It wants its own PR - and that PR should be the one to decide whether `clang-tidy` stays `continue-on-error`, since that flag is what let this hide.

### Also in this push

`main` moved (#1966) and the PR went un-mergeable; `origin/main` is merged in and the `CHANGELOG` conflict resolved by keeping both entries. `git diff origin/main --stat` is back to the same two files.

---
_Generated by [Claude Code](https://claude.ai/code)_

kekzl:
Closing the loop on "the number to read is the one after this" - and on why `clang-tidy` ran in the first run when it should not have.

It ran because the change detection had never worked. The `Build` step was the first git caller in the job and `safe.directory` was set 50 lines later, so `git cat-file` died on the ownership refusal with stderr suppressed and the fail-open returned `code=true` every run. That also means the docs-only skip had never fired once. Fixed in e4a3ac7; description updated with the evidence.

Verified on the current head:

```
non-docs changes: code=true (src|tools .cpp: false)
changed files:
.claude/skills/building-and-testing/SKILL.md
.claude/skills/shipping-prs/SKILL.md
.github/workflows/ci.yml
CHANGELOG.md
```

Populated list, `cpp: false`, no fall-open warning, `clang-tidy` skipped.

| job | before (median, 8 runs) | after |
|---|---|---|
| **whole run** | **~19 min** | **5.8 min** |
| Sanitizers | 19.0 | **2.6** |
| Build | 6.2 | 3.9 |
| clang-tidy | 1.8 | skipped |

Splitting the Sanitizers effect: cold ccache with the single gencode was 13.7 min, warm 2.6-3.1 min. So roughly a third of that lane's saving is the dropped `compute_120f` gencode and the rest is the cache.

CI green, mergeable, still a draft.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #2062, opened 2026-09-21
docs(plans): size the CPU-side expert path before writing it (13 % headroom, not 2x)

kekzl:
Sizes the last remaining decode lever before anyone writes it, and ships the probe.

The intuition behind a CPU-side expert path is that host RAM is much faster than the link, so the missed experts should be computed where they already sit instead of being shipped to the card. On this box that intuition is wrong by a wide margin.

`tools/analysis/cpu_expert_gemv_probe.cpp` reads random 0.88 MiB experts out of a 4 GiB slab, 9800X3D, 16 threads, 537 misses per token (the 62.7 % hit rate measured in #2061):

| | ms/token | GB/s |
|---|---|---|
| read only, memory-bound ceiling | 8.25 | 60.0 |
| dequant + GEMV, scalar | 28.61 | 17.3 |
| the same bytes over PCIe | 9.52 | 52.0 |

Host RAM for scattered expert-sized blocks is 58-60 GB/s, 13 % above the link. A perfect AVX-512 NVFP4 GEMV sitting exactly on that roofline would save 1.2 ms of 9.5 on that component, and the scalar version is 3.5x off it. Eight threads reach the same read bandwidth as sixteen, so the ceiling is the memory controller, not core count.

That does not kill the idea, it repositions it: the case for a CPU path is overlap, the CPU computing its misses while the GPU computes its hits, and it has to be written against an 8.3 ms floor rather than an imagined 4 ms one.

Docs and one probe, no engine code. `pre-push: no source changes, skipping verify`.

https://claude.ai/code/session_01Tzmv6adUksv8Mcjtsr26i5


kekzl:
Branched off the previous topic branch instead of fresh origin/main, so the squash merge of #2061 made it conflict. Same content, reopened cleanly as the next PR.

